PR #4, opened 2015-02-18
CoreAudioAPI extending with demo

KvanTTT:
Unfortunately can not check Windows Store assembly WACK.
Changes not break any functionality in existing NAudio code.

PR #5, opened 2015-02-26
Expose IAudioStreamVolume from WsapiOut, and AudioClient.

rassilon:
This exposes `IAudioStreamVolume` from `AudioClient/WsapiOut` as `AudioStreamVolume` object.

`IAudioStreamVolume` is claimed to work on Vista or later.

I wasn't quite sure how to go about writing up a test for this. Thoughts? I tested all of the methods by creating a sample app that played 3 MP3 files simultaneously at different `IAudioStreamVolumes`.
As well as testing the `GetAllVolumes` and `SetAllVolumes` worked correctly.

Fyi,
Bill

### DIFF
--- a/NAudio/CoreAudioApi/AudioClient.cs
+++ b/NAudio/CoreAudioApi/AudioClient.cs
@@ -15,6 +15,7 @@ namespace NAudio.CoreAudioApi
         private AudioRenderClient audioRenderClient;
         private AudioCaptureClient audioCaptureClient;
         private AudioClockClient audioClockClient;
+        private AudioStreamVolume audioStreamVolume;
 
         internal AudioClient(IAudioClient audioClientInterface)
         {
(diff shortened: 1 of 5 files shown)


markheath:
thanks for this. Will need to check out that this doesn't end up clashing with another [pull request](https://github.com/naudio/NAudio/pull/4) that is doing the same thing, but not quite ready yet to be merged in

rassilon:
I'm not sure how this could conflict with the other since we're not doing the same thing. They are both volume related prs, but at different levels in the audio stack. 

Kinds of volume interfaces:
Channel: this can control the volume for only one channel of an AudioClient.  ie left but not right
Audio
AudioStream: this controls the volume for all channels of an AudioClient. Ie right and left
This is what this pr addresses. 
Session volume: session volume is the volume level of all AudioClients in the session.  This is what I think the other pr addresses. Example : if an app creates 5 wasapi outs they each have their own audio stream volume but share their session volume. 
Endpoint volume: the volume level for the audio Endpoint. Naudio already supports this. 

Bill

markheath:
yeah, sorry my mistake. Will try to get a code review done on this soon.

PR #10, opened 2015-03-23
Add sample-exact seeking to Mp3FileReader

protyposis:
This changeset adds frame-exact seek capabilities to the Mp3FileReader as explained in [this blogpost](http://mark-dot-net.blogspot.co.at/2010/11/state-of-mp3-playback-support-in-naudio.html). It also fixes frame-granular repositioning, position reporting, and handles the decoder warmup to return correct data blocks after a seek (instead of starting attenuated when the overlap of the previous frame is missing). There's comments in the code explaining the most important changes. Now all of that works as expected, but because of these changes, external code written to work around these issues may now lead to wrong results (but that's just a side note, it does not concern NAudio itself).

Btw thanks for this nice library, I've been using it a lot and now I can finally contribute something back :)

### DIFF
--- a/NAudio/Wave/WaveStreams/Mp3FileReader.cs
+++ b/NAudio/Wave/WaveStreams/Mp3FileReader.cs
@@ -330,7 +330,7 @@ namespace NAudio.Wave
                     Mp3Index mp3Index = null;
                     for (int index = 0; index < tableOfContents.Count; index++)
                     {
-                        if (tableOfContents[index].SamplePosition >= samplePosition)
+                        if (tableOfContents[index].SamplePosition + tableOfContents[index].SampleCount > samplePosition)
                         {
                             mp3Index = tableOfContents[index];
                             tocIndex = index;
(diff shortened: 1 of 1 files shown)


markheath:
thanks, looks good. I will try to do a full code review soon. What testing have you been able to perform? Have you checked that it still works if an alternative MP3 Frame decoder is plugged in?

protyposis:
The changes have been made without taking any strong assumptions about the decoder so it should be compatible with different decoders, and I validated it working with the DMO decoder. The only assumption currently taken is that a decoder does not hold back more than one frame, thus does not return more than two decoded frames at the same time, but that constraint was coded into the reader anyway (decompressBuffer.Length == two frames).

I obviously didn't add unit tests as they would not make sense without supplying an MP3 test file, and even then it depends on the decoder implementation which exact data it returns so validating on byte level if the decoded data equals the expected data isn't trivial. I am however using the MP3 reader in a graphical multitrack audio data analysis application (not published yet) that heavily relies on correct positioning and random seeks, which is how I identified the problems in the first place and also validated them being solved. I basically manually compared graphical representations of the sample series of decoded MP3s with their WAV source under various circumstances, and my visual testing tells me that this changeset works correctly.

markheath:
great thanks, will get it merged

protyposis:
That was fast, thanks!

PR #11, opened 2015-03-24
Replace SynchronizationContext with Dispatcher for WPF applications

thorstenbaek:
To make NAudio WaveIn work for our WPF with FunctionCalllback I needed to replace the SynchronizationContext with Dispatcher for WPF applications.
My solution is based on this article:
http://stackoverflow.com/questions/1949789/using-synchronizationcontext-for-sending-events-back-to-the-ui-for-winforms-or-w

The static variable Application.Current is null for non-WPF applications. I tested the waveIn recorder demo in NAudioDemo to verify that this code runs for WinForms applications.

Best Regards
Thor Stenbæk
DIPS ASA

### DIFF
--- a/NAudio/NAudio.csproj
+++ b/NAudio/NAudio.csproj
@@ -67,11 +67,13 @@
     <DocumentationFile>bin\Release\NAudio.XML</DocumentationFile>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="PresentationFramework" />
     <Reference Include="System" />
     <Reference Include="System.Data" />
     <Reference Include="System.Drawing" />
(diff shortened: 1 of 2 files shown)


thorstenbaek:
Sorry! I still get dead lock situations. I close the request and try again later...

markheath:
I strongly recommend using WaveInEvent instead of Function callbacks. Function callbacks have caused me no end of headaches and I don't use them anymore.

thorstenbaek:
Yes, I just discovered the WaveInEvent class. It solved my problems perfectly. 

Thank you very much for a great Audio Api in .Net.

Thor Stenbæk 

Sendt fra min iPhone

PR #13, opened 2015-04-17
Prevent Unhandled Exception from async void StartRecording

kamenlitchev:
In Windows 8 whenever the library did not have access to the Microphone, any call to StartRecording will succeed due to the method being "async void". However, the capture thread would throw System.UnauthorizedAccessException which will then in turn crash the whole application as Unhandled Exception.

With this change StartRecording is still done as async method, but any exceptions in Activate() are propagated to the caller.

Not the prettier of code, but it works. Note that DoRecording() is still done on another thread.

Code reference:
Exception Handling (Task Parallel Library): https://msdn.microsoft.com/en-us/library/dd997415(v=vs.110).aspx
Best Practices in Asynchronous Programming: https://msdn.microsoft.com/en-us/magazine/jj991977.aspx

### DIFF
--- a/NAudio.Win8/Wave/WaveInputs/WasapiCaptureRT.cs
+++ b/NAudio.Win8/Wave/WaveInputs/WasapiCaptureRT.cs
@@ -157,20 +157,24 @@ namespace NAudio.Wave
         /// <summary>
         /// Start Recording
         /// </summary>
-        public async void StartRecording()
+        public void StartRecording()
         {
             this.stop = false;
-            var audioClient = await Activate();
-            Task.Run(() => CaptureThread(audioClient));
-            
-            /*Task.Run(
-                async () =>
-                    {
-                        var audioClient = await Activate();
-                        //InitializeCaptureDevice(audioClient); - now done in the activate callback
-                        CaptureThread(audioClient);
-                    });*/
 
+            try
+            {
+                Task.Run(new Func<Task>(async () =>
+                {
+                    AudioClient audioClient = await Activate();
+                    Task.Run(() => DoRecording(audioClient));
+                })).Wait();
+            }
+            catch (AggregateException ae)
+            {
+                Debug.WriteLine("Error starting thread: " + ae.InnerException);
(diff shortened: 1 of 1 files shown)


kamenlitchev:
To reproduce the described behaviour:
1. On Windows 8(.1) start the NAudioWin8Demo
2. Press Record button
3. On Microphone access prompt choose Block (if app already has access, use PC Settings -> Privacy -> Microphone to revoke it
4. Observe Unhandled crash that closes the whole app

With the proposed change NAudio users can wrap WinCaptureRT.StartRecording in try/catch and can handle the exception.

markheath:
could this exception not have been caught by using ``await recorder.StartRecording()`` in the ``MainPageViewModel`` instead?

kamenlitchev:
No. "async void" methods (as StarRecording in master) do not return exceptions to the caller. "async void" methods also cannot be awaited in the intended "await" sense, since they do not return to the caller any result and exceptions. Awaiting async void method is the same as starting new thread and not waiting it.

If you want StartRecording to be properly awaited and exceptions from activations propagated to the caller, the signature must be changed to Task StartRecording which means change in the implemented interface. Alternatively, there can be additional Task StartRecordingAsync method which properly propagate exceptions to the caller.

See http://www.jaylee.org/post/2012/07/08/c-sharp-async-tips-and-tricks-part-2-async-void.aspx where the samples and explanation are quite good: "async void methods exist for the single purpose of being used as an event handler. More specifically, it has been introduced to support UI elements event handlers, nothing else."

See: https://msdn.microsoft.com/en-us/magazine/jj991977.aspx

markheath:
Yes, my mistake, I was thinking it was already returning a task, but we can't do that without changing the IWaveIn interface. Let's go with your fix for now, but I am still wondering if it is better to redesign the public interface to fit in better with an async programming model for the future.

kamenlitchev:
Thanks :) A redesign will brake clients and I'm not sure if people will like it. Maybe new Task ...Async method with Obsolete of the void method will be OK, though.

PR #14, opened 2015-04-21
Dispose of locally created stream in WaveFileReader when there is an exception thrown in the constructor

adraut:
When creating a WaveFileReader the ChunkedReader can throw exceptions.  This is OK when a stream is passed in because the WaveFileReader does not own the stream.  But when a file path is passed in the WaveFileReader does own the stream, and does not properly dispose of it (dispose is not called on objects that failed to construct) leaving it locked by the process running NAudio.  This fixes that behavior.


markheath:
great, looks good

PR #15, opened 2015-04-23
Email change

weitzhandler:
Changed mark's email to avoid him getting spammed...


rogersachan:
I don't think this pull request is necessary anymore

markheath:
yes thanks for the sentiments, but it too late to save me from spam!

PR #23, opened 2015-06-23
Windows Universal Project

kamenlitchev:
Project for Windows Universal apps


markheath:
hi, Just have a few quesions
Have you tested that the Win 8 project still works on Windows 8? I've not come across the DLLs you are referencing before e.g. api-ms-win-core-handle-l1-1-0.dll
Are these safe to use in Win 8/8.1? Is this the recommended way to access these APIs now, or is this some undocumented way to get at them?

kamenlitchev:
Both my PCs are Windows 10. I have a Windows 8.1 one but did not manage to test on it. Will try it this weekend.

However, according to the official documentation for Windows 8, these DLLs are the ones to be used. Check the Windows 8 API Sets in MSDN: https://msdn.microsoft.com/en-us/library/windows/desktop/dn505783(v=vs.85).aspx

PR #24, opened 2015-06-26
ArgumentOutOfRangeException in CircularBuffer.cs

roadz99:
Make CircularBuffer thread safe when calling reset while the buffer is still in use.

### DIFF
--- a/NAudio/Utils/CircularBuffer.cs
+++ b/NAudio/Utils/CircularBuffer.cs
@@ -117,12 +117,12 @@ namespace NAudio.Utils
         /// </summary>
         public void Reset()
         {
-			lock (lockObject)
-			{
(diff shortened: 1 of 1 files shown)


KvanTTT:
Deadlock?

Actions sequence:
* Call ```Advance```.
* Locking ```lockObject```.
* Call ```Reset``` method inside ```Advance```, but waiting for the lock releasing.

roadz99:
If the thread has already acquired the lock in the Advance method then the second call to acquire the lock in the Reset method will succeed because the thread already owns the lock.
I can't see any circumstance where adding extra threads could cause a deadlock - the methods I've changed are extremely short and don't call out to anything external.

markheath:
I think this is resolved by #55

PR #25, opened 2015-06-30
Cleaned NAudioWpfDemo using directives and corrected summaries

MichalisN:
Code cleanup mostly.
Removed unused using directives, corrected some summaries and changed the return condition at CanExecute method in RelayCommand.cs file for better readability.

### DIFF
--- a/NAudioWpfDemo/App.xaml.cs
+++ b/NAudioWpfDemo/App.xaml.cs
@@ -1,10 +1,7 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.Configuration;
-using System.Data;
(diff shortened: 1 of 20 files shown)


markheath:
hi MichalisN,
I am happy to accept this pull request, as it relates to the WPF demo app only, and makes only minimal changes to actual code. But just to let you know that I usually reject pull requests that attempt to clean up the whole library using a tool like ReSharper, due to risk of inadvertently introducing breaking changes. I code review all pull requests, but when hundreds of changes are made in one go, it is not easy for me to have confidence that there are no regressions.

PR #28, opened 2015-07-11
RawSouceWaveStream Minor Update

danm36:
Added the capability for the RawSourceWaveStream to be initialized by a byte array.
Fixed position setter to prevent potential runtime playback corruption when seeking.


markheath:
nice, although I'd usually do the block align by doing value - (value % waveFormat.BlockAlign). the decimals and rounding seems a little overkill to me

danm36:
Ah, that would actually make a lot more sense (I was a little tired when I make the change first, and was just happy I got a result). Updated.

markheath:
thanks

PR #34, opened 2015-07-31
Do not lock the file when loading it using the WaveFileReader

SimonSimCity:
I am currently writing an application that works with WAVE files, where another app continuously adds data to that file.

Today I discovered, that the `WaveFileReader` is calling `File.OpenRead(waveFile)` which is the same as calling `File.Open(waveFile, FileMode.Open, FileAccess.Read, FileShare.Read)`. As you can hereby see, this handler will block the file for writing. The operation is aborted and the system fires an exception, telling you that another process currently uses this file.

The solution here was to replace it by an explicit call of `File.Open()` and adjust the last parameter to be `FileShare.ReadWrite`, saying that other applications are allowed to read and write to that file while it's opened by that handler.

As this is an instance, called `WaveFileReader`, I don't see any problem adding this feature here. This handler can't write anyways and I don't see a reason to deny that.

### DIFF
--- a/NAudio/Wave/WaveStreams/WaveFileReader.cs
+++ b/NAudio/Wave/WaveStreams/WaveFileReader.cs
@@ -30,7 +30,7 @@ namespace NAudio.Wave
         /// fix this reader to support it
         /// </remarks>
         public WaveFileReader(String waveFile) :
-            this(File.OpenRead(waveFile), true)
+            this(File.Open(waveFile, FileMode.Open, FileAccess.Read, FileShare.ReadWrite), true)
         {            
         }
 


markheath:
this fix will sort of work, but remember that the WAV file format includes length values in the header, which will already have been read by WaveFileReader, so attempting to write a WAV file while you play it is not likely to work very well.

SimonSimCity:
I had some probs with that one here, @markheath. I did some intense testing on that one yesterday.

We use Pro Tools 9 (http://www.avid.com/us/products/family/pro-tools) to record the audio. As Pro Tools always already at the start of the recording, plans on a 3h file, it reserves the space on the harddisk at once.

The problem we encounted now is, that if your file is written to the file-system NTFS, the file is allocated as to be around 300mb - in our example. As I don't know the marker of where Pro Tools currently is, I can open the file-handler and can get off to a state, where Pro Tools is going to write something (I am ahead of the current recording). In this case, I'd get the bytes that are on the harddisk before (in our case an old - deleted WAVE file - just random). So, this is quite critical.

To cut the rest of the story, we ended up with having Pro Tools on a Mac (since the file system of a Mac supports dynamic files without locking) and opened a share using Samba to make the file accessible for my application.

So, as this is quite a complex setup - and it only works in certain circumstances (due to some Samba-behavior we have to use OS X Mavericks and can't update if we don't want to break that behavior), you can just skip that PR.

Just to have it documented: We got something working, but it has to do with quite some low-level facts like locking of dynamic files having a dynamic file size and how your OS or your file system handles that.

markheath:
OK great, will leave it for now. I've written code myself in the past to play from files that were being written, but I did so with my own customised versions of `WaveFileReader` and `WaveFileWriter`. Glad you found a solution in the end anyway.

PR #41, opened 2015-08-30
Fix fatal error on WasapiCaptureRT when starting record

Kassoul:
It's because ActivateAudioInterfaceAsync must be called from UI Thread.
Also added few simplifications on WasapiCaptureRT

### DIFF
--- a/NAudio.Win8/Wave/WaveInputs/WasapiCaptureRT.cs
+++ b/NAudio.Win8/Wave/WaveInputs/WasapiCaptureRT.cs
@@ -20,6 +20,7 @@ namespace NAudio.Wave
     /// </summary>
     public class WasapiCaptureRT : IWaveIn
     {
+        static readonly Guid IID_IAudioClient2 = new Guid("726778CD-F60A-4eda-82DE-E47610CD78AA");
         private const long REFTIMES_PER_SEC = 10000000;
         private const long REFTIMES_PER_MILLISEC = 10000;
         private volatile bool stop;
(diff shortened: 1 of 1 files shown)


markheath:
hi, thanks for this. Can you tell me a bit more about what system you are seeing this error on, and what you have tested it on? Did it not work for you in the NAudio WinRT demo app? And have you tested your changes still work with the NAudio WinRT demo app?

Kassoul:
Hi Mark,
I have seen this error on my Windows 8.1 64 machine, on my software and also on the NAudio WinRT demo app.
Of course, I have tested my changes on the NAudio WinRT demo app and recording now works.
Thanks for merging.
Philippe.

markheath:
great, thanks. I had a chance to test it out on Win 10 myself and reproduced the issue and resolution. I'm sure it used to work, so don't know why it has just broken. Really I suppose this requires the `StartRecording` method to be made `async`, but this requires a modified version of `IWaveIn`. Might consider that for an NAudio vNext.

kamenlitchev:
Guys, with this commit we are hitting the "async void" trap again that was fixed with my commit. This means that any error thrown during StartRecroding will get swallowed and clients of the library will have no clue what is going on.

Better thing either of interface change "async Task" or wrap all async code as synchronous one and call ActivateAudioInterfaceAsync outside any async calls

kamenlitchev:
How about:

```
try  
{  
    var icbh = new ActivateAudioInterfaceCompletionHandler(
	    ac2 =>{ InitializeCaptureDevice((IAudioClient)ac2);}); 
     
    IActivateAudioInterfaceAsyncOperation activationOperation; 
    // must be called on UI thread 
    NativeMethods.ActivateAudioInterfaceAsync(
	    device, IID_IAudioClient2, IntPtr.Zero, icbh, out activationOperation);
    Task.Run(new Func<Task>(async () =>  
    {  
        var audioClient2 = await icbh; 
		
		// We do not need to await this one
        Task.Run(() => DoRecording(new AudioClient((IAudioClient)audioClient2))); 
    })).Wait();  
}  
catch (AggregateException ae)  
{  
    Debug.WriteLine("Error starting thread: " + ae.InnerException);  
    throw ae.InnerException;  
}  
```

Here we have:
- Activate on UI thread
- Catching of other async errors
- No "async void" calls
- The DoRecording call is not awaited

markheath:
we could try this. I'm thinking that for NAudio 1.8 we could introduce a breaking change and make the `IWaveIn` interface have `StartRecordingAsync` returning a Task

PR #55, opened 2015-11-10
Synchronized byteCount variable in CircularBuffer

dmitry-ra:
byteCount variable changed simultaneously from a different threads. That is it should be synchronized.
This PR will close issue 54: naudio/NAudio/issues/54


dmitry-ra:
Reference naudio/NAudio/pull/24
/cc @roadz99

Issue #59, opened 2015-11-25
Windows 10 Mobile ole32.dll error when playing


juanpaexpedite:
Hello, I suppose you know about this issue, tested on the latest release of Mobile:

Unable to load DLL 'ole32.dll': The specified module could not be found. (Exception from HRESULT: 0x8007007E):
When play a file,  variant.Clear(); calls to 

public void Clear()
        {
            PropVariantClear(ref this);
        }

        [DllImport("ole32.dll")]
        private static extern int PropVariantClear(ref PropVariant pvar);

what is not available on Mobile, I understand.

czesiu:
I have the same error on Raspberry Pi on Windows Universal Sample.

Here is the full stack trace:
    
    Unable to load DLL 'ole32.dll': The specified module could not be found. (Exception from HRESULT: 0x8007007E)
    at NAudio.CoreAudioApi.Interfaces.PropVariant.PropVariantClear(PropVariant& pvar)
    at NAudio.CoreAudioApi.Interfaces.PropVariant.Clear()
    at NAudio.Wave.MediaFoundationReader.GetLength(IMFSourceReader reader)
    at NAudio.Wave.MediaFoundationReader..ctor(String file, MediaFoundationReaderSettings settings)
    at NAudioUniversalDemo.MediaFoundationReaderUniversal..ctor(MediaFoundationReaderUniversalSettings settings)
    at NAudioUniversalDemo.MediaFoundationReaderUniversal..ctor(IRandomAccessStream stream)
    at NAudioUniversalDemo.MainPageViewModel.CreateReader()
    at NAudio.Win8.Wave.WaveOutputs.WasapiOutRT.Init()
    at NAudio.Win8.Wave.WaveOutputs.WasapiOutRT.<PlayThread>d__25.MoveNext()```

kamenlitchev:
Seems like another DLL reference that needs to be changes. As per https://msdn.microsoft.com/en-us/library/windows/desktop/hh802935(v=vs.85).aspx#api_sets_available_in_windows_10 the PropVariantClear seems to be placed in "api-ms-win-core-com-l1-1-1.dll"

juanpaexpedite:
kamenlitchev You are totally right! changing to:

[DllImport("api-ms-win-core-com-l1-1-1.dll")]
private static extern int PropVariantClear(ref PropVariant pvar);

works both on mobile and desktop, 
czesiu  would you try on Raspberry Pi?

czesiu:
[...]

PR #61, opened 2015-12-01
 [Issue #59] Windows 10 Mobile ole32.dll error when playing

kamenlitchev:
Fix errors for native call of PropVariantClear reported in #59 . Also up Windows 10 Demo max tested version to latest Windows 10 SDK (1511).


kamenlitchev:
Note - PropVariantClear for Windows 8 is changed to api-ms-win-core-com-l1-1-0.dll while Windows 8.1 should use api-ms-win-core-com-l1-1-1.dll

Not sure if it's not better to leave the dll import to ole32.dll instead for Windows 8 and implement the change

markheath:
soory for delay, will hopefully get round to testing this soon. I've got some more UWP fixes of my own I need to merge this with

kamenlitchev:
Thanks :) Shall I return the original "ole32.dll" for Windows 8? And leave the change only for Windows 10?

markheath:
I have to confess I'm not sure what the best option is. Does `[DllImport("api-ms-win-core-com-l1-1-0.dll")]` only work on 8 or can it work on 8.1 too? To be honest the UWP is probably the most important thing going forwards, but if it's not too hard to keep the Win8 one working on all platforms then that would be a good idea.

kamenlitchev:
Done. I have reverted Win8 to include the .NET PropVariantNative "ole32.dll" reference. Just to be on the safe side.

Issue #64, opened 2015-12-17
AudioSessionControl Dispose and UnRegisterEventClient throws COMException


Drapondur:
Both methods AudioSessionControl.Dispose and UnRegisterEventClient are throwing a ComException under the following circumstances
1. Call AudioSessionControl.UnRegisterEventClient 
2. Then call AudioSessionControl.Dispose
Both methods are checking if the registered audioSessionEventCallback is null prior to the IAudioSessionControl.UnregisterAudioSessionNotification call but are not setting the audioSessionEventCallback member to null. Thus if UnRegisterEventClient had unregistered the event client then Dispose (and the finalizer) are calling IAudioSessionControl.UnregisterAudioSessionNotification again which then causes the exception.

This happened running on on Win 7 64 bit.

PR #65, opened 2015-12-24
Add Loopback flag with EventCallback

MathewSachin:
1. When using WasapiLoopbackCapture, it is necessary to add AudioClientStreamFlags.Loopback flag.
But, in the class WasapiCapture, if EventCallback is used, then this flag is not added.
2. Also #64 has been fixed


MathewSachin:
Travis CI builds were already failing.
Please ignore the CI check that failed

PR #79, opened 2016-02-27
Fix issue I had with exclusive WasapiOut to a 7.1 sound card.

mikelehen:
When trying to use WasapiOut in exclusive mode with my 7.1 sound card, it would revert to the "Last Last Last Chance (Thanks WASAPI)" case and use 2 channel 16-bit output, which isn't what I wanted.

Furthermore, I found that any 8 channel WaveFormatExtensible that I created was rejected by IsFormatSupported().  I tried different sample rates and different bitrates, but nothing worked.

Eventually I found that you can ask the device what format to use via the PKEY_AudioEngine_DeviceFormat property.  So I added code to WasapiOut to do that, and it worked!  This led me to the discovery that it wanted a dwChannelMask of 0x63F instead of 0xFF.

So I fixed both problems.  I'm not 100% sure these fixes are completely correct (because I have very little experience with these APIs), but either fix is sufficient to fix my specific problem.  I'd love if one or both could be incorporated into NAudio.

### DIFF
--- a/NAudio/Wave/WaveOutputs/WasapiOut.cs
+++ b/NAudio/Wave/WaveOutputs/WasapiOut.cs
@@ -325,6 +325,25 @@ namespace NAudio.Wave
                             correctSampleRateFormat = null;
                         }
 
+                        // If still null and we're using exclusive mode, try to get the device format property.
+                        if (correctSampleRateFormat == null && shareMode == AudioClientShareMode.Exclusive)
+                        {
+                            // Based on https://stackoverflow.com/questions/22616924/wasapi-choosing-a-wave-format-for-exclusive-output
+                            byte[] waveFormatBytes = (byte[])mmDevice.Properties[PropertyKeys.PKEY_AudioEngine_DeviceFormat].Value;
+                            if (waveFormatBytes != null)
+                            {
+                                GCHandle handle = GCHandle.Alloc(waveFormatBytes, GCHandleType.Pinned);
+                                try
+                                {
+                                    correctSampleRateFormat = (WaveFormatExtensible)Marshal.PtrToStructure(handle.AddrOfPinnedObject(), typeof(WaveFormatExtensible));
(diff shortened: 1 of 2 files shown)


NAudio/Wave/WaveOutputs/WasapiOut.cs
```diff
@@ -325,6 +325,25 @@ public void Init(IWaveProvider waveProvider)
                             correctSampleRateFormat = null;
                         }
 
+                        // If still null and we're using exclusive mode, try to get the device format property.
+                        if (correctSampleRateFormat == null && shareMode == AudioClientShareMode.Exclusive)
+                        {
+                            // Based on https://stackoverflow.com/questions/22616924/wasapi-choosing-a-wave-format-for-exclusive-output
+                            byte[] waveFormatBytes = (byte[])mmDevice.Properties[PropertyKeys.PKEY_AudioEngine_DeviceFormat].Value;
+                            if (waveFormatBytes != null)
```
mikelehen: I'm not sure if this code is right...

1. Is there an easier way to marshal a byte[] into a WaveFormatExtensible?
2. Is this even correct?  I think PKEY_AudioEngine_DeviceFormat actually retrieves a WAVEFORMATEX, not a WAVEFORMATEXTENSIBLE.  So this may be wrong.

mikelehen: Well, actually I'm confused.  https://msdn.microsoft.com/en-us/library/windows/desktop/dd316580(v=vs.85).aspx says it's a WAVEFORMATEX, but the dwChannelMask is getting set correctly, which I think means (in my case at least) it must be a WAVEFORMATEXTENSIBLE.

Maybe I need to take a look at the byte count to decide or something.  Let me know if you have any ideas. :-)

To be clear, the code does work for me as-is. *shrug*

Issue #93, opened 2016-05-21
NAudio needs support for asynchronicity (and probably .NET 4.0 and 4.5)


kkm000:
.NET 4.0 added asynchronous operations to the definition of `System.IO.Stream`, and 4.5 took that a step further by providing future-promise asynchronous model with `Task *Async()` methods. NAudio currently targets .NET 3.5. This version did have a basic support for asynch operation (`Begin{Read,Write}`, although harder to use than the new style); but NAudio does not support even that. This indeed creates a problem on servers, like in my case when I am using an instance of  `WebResponseStream` to construct a `WaveFileReader`. The latter blocks when reading data from the underlying stream, and that requires a complex thread management in a service program (I cannot block in a thread pool thread, since they are scarce).

NAudio is an amazing library, and would be awesome to keep it up with the way current architectures are developing.

dlnieves:
Totally agree. Think of game development also!!!!

kkm000:
@markheath: Is there interest in this at all? Looks like a profound change, as it would require building and packaging multiple versions of the library, akin to what is done now with separate projects in `NAudio.Win8` and `NAudio.Universal` (I would use build targets rather than creating 2 more projects but your choice). I would start hacking it, but I am not very gung ho about maintaining a fork. What are your thoughts?

markheath:
yeah, a lot of NAudio was written back in the days of .NET 1.0. Some parts would be hard to make async, but for things like WaveFileReader / Writer it would be a nice addition.

kkm000:
@markheath, I'm taking a liberty to hijack this space for questions on the NAudio build architecture. There is a feature in the FAKE script that looks sloppy at the first sight. The Debug build is unit tested, but the Release build is packaged. Is this a bug, or intended for a reason I do not comprehend?

```f#
"Clean" 
    ==> "DebugBuild"
    ==> "Test"
    ?=> "ReleaseBuild"
    ==> "Release"
```

kkm000:
Ahem? PTAL.

markheath:
yes, I run the unit tests on the debug build but the nuget package contains the release build. Is there a problem with that?

kkm000:
No, not a problem as it is, just looks strange indeed. I could not understand what is the point of building a separate and (sometimes unintentionally) different configuration for unit testing only. I just want to understand the benefits.

[...]

PR #99, opened 2016-06-14
[WIP don't merge] Supporting .NET 4.0 and 4.5

kkm000:
@markheath, this is referring to #93. First thing first, I need to compile at least 3 different versions of NAudio (original .NET 3.5/CLR 2.0; unchanged .NET 4.0/CLR 4.0; and one with async support added for .NET 4.5 - this is where the actual feature could only be added). 

I am hesitating as to what is the best way to support multiple build targets. Here's a stab at a single solution with multiple build configurations (Release, Release40 and Release45, Debugs same 3). Personally, I always prefer this, since e. g. added source file is automatically propagated to all builds.

However, this conflicts somewhat with the existing FAKE build script, that expects only Release and Debug targets in all 3 solutions, the other 2 being Win8 and Universal.

Now, I see 3 paths to reconcile the build.

1. Continue the old approach and duplicate solutions and projects twice more, for .NET 4.0 and 4.5, bumping the number of .sln files to maintain up to 5.

2. Use my approach to integrate Win8/Uni builds into single solution and single project per library/app.

3. Use my approach only for .NET 3.5/4.0/4.5, and leave the other 2 alone. 

Now, to me (1) looks icky, as project maintenance seems to slip out of control. (2) seems right but requires a lot of work before I can even get to implementing the real McCoy, and (3) is somewhat half-assed, and requires adding an edge case to the FAKE script.

If I were to take the decision, I'd start with (3) with the goal to land eventually at (2). What do you think?


markheath:
Hi there, at the moment I am a little bit reluctant to add multiple additional versions of NAudio. The current build works just fine in .NET 4 and .NET 4.5. If we did, then I would prefer option 3 to option 1. I'm not sure I fully understand option 2, but I suspect that that UWP is going to be different enough to not share a common solution with the main library. I am currently researching the UWP AudioGraph API which may feed into some more work on the Win 10 project in the future.

kkm000:
I'll maintain own fork then, thanks.

kkm000:
Ah, and you might want to close #93 please with a comment like "not going to happen" then.

markheath:
hi @kkm000, I wasn't meaning to say "not going to happen". The main reason behind my reluctance is the increased workload of testing and maintaining multiple versions. I'm hoping that I can expand the NAudio team in the future to make things like this more managable.

kkm000:
> I wasn't meaning to say "not going to happen".

Well, I must have misunderstood you. You said that “The current build works just fine in .NET 4 and .NET 4.5.” when it is in fact does not, as reported in #93. So I assumed, probably incorrectly, that you do not agree this is a problem. Mind you, it's not a “nice to have” feature--pumping streams in threads is a real bottleneck.

> I'm hoping that I can expand the NAudio team in the future to make things like this more managable.

Looks like I am missing something again. I thought NAudio was a one-man project, not any larger company, is it? If it is, then it does not compute when I offered exactly what you seem to be talking about, namely, my work for this project. I do not really understand.

>I'm not sure I fully understand option 2, but I suspect that that UWP is going to be different enough to not share a common solution with the main library.

I should look again into the project structure, but it seemed to me that they still share most of it. System-specific files could use, for example, `#if HAVE_UWP`/`#endif` blocks, and then the constants can be defined my multiple project configurations, say, `Release-4.0`, `Release-SL5`, `Debug-10` etc etc. An example  of MS' own solution building a dozen different configurations: https://github.com/Reactive-Extensions/Rx.NET/blob/v2.2.5/Rx.NET/Source/Rx.sln#L56

PR #107, opened 2016-07-08
new StreamMediaFoundationReader which takes a stream instead of a file

softlion:
This opens new frontiers to NAudio, as it can now read and decode any online streams successfully (no other method works). Well, only fully read compressed streams (ie: memorystreams containing frames of compressed audio)

### DIFF
--- a/NAudio/Wave/WaveStreams/ComStream.cs
+++ b/NAudio/Wave/WaveStreams/ComStream.cs
@@ -1,0 +1,160 @@
+using System;
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Runtime.InteropServices.ComTypes;
+
+namespace NAudio.Wave
+{
+    public class ComStream : Stream, IStream
+    {
+        private Stream stream;
+
+        public override bool CanRead => stream.CanRead;
+
+        public override bool CanSeek => stream.CanSeek;
+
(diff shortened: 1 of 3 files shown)


markheath:
awesome! I tried a few times to do this but never with any success. Have a bit of a hectic schedule at the moment, but hoping to give this a try soon and get it merged in.

softlion:
I tried a modified version of this code (because your code has missing default constructor) and it works well. The trick is the ComSteam class that comes from the cscore github project.

markheath:
ok, might be good just to double-check with Florian that they are OK with you bringing that class into NAudio

softlion:
It is the same licence (Ms-PL) so there is no need to ask.

markheath:
ok, I've reached out to Florian and not heard back from him so let's go for it. I wonder if this would be better if it was just another constructor of `MediaFoundationReader` though? What are your thoughts?

softlion:
Because of the way the constructor of this class works (it does not supports null urls), it seems not possible without some refactoring. I'm not deep in this project, so I can't tell the impacts. That's why i decided to do it this way. Feel free to add a new constructor !

markheath:
OK cool, I'll have some time in about a week to process some NAudio pull requests so I plan to merge this one in then

PR #117, opened 2016-09-06
Adding MidiEvent.Clone()

### LINKED ISSUE #114: Request for MidiEvent.Clone
I want to repeat a section of a MIDI file and right now I don't see a way to do that. It would be perfect to have an abstract `Clone` method on the MidiEvent class, just like Stephen Toub did [here](https://github.com/stephentoub/MidiSharp/blob/54c5f4edafa5e0ac8110faa25fdea6a395e6f0c9/MidiSharp/Events/MidiEvent.cs#L223).

I'm willing to submit a PR as well.

jnm2:
There's a couple of things in the inheritance hierarchy that could optionally be optimized that I'll comment on in the source.

Closes #114.


NAudio/Midi/MidiEvent.cs
```diff
@@ -192,6 +192,13 @@ public MidiEvent(long absoluteTime, int channel, MidiCommandCode commandCode)
         }
 
         /// <summary>
+        /// Creates a deep clone of this MIDI event.
+        /// </summary>
+        public virtual MidiEvent Clone() => (MidiEvent)MemberwiseClone();
```
jnm2: I'd prefer to make this abstract so that the onus is always on the subclass to figure out if it's doing the right thing to clone. That would require making MidiEvent abstract which would require either introducing an UnknownMidiEvent or making sure all midi events are accounted for or throwing an exception on unknown event types.

NAudio/Midi/MetaEvent.cs
```diff
@@ -44,6 +44,11 @@ public MetaEvent(MetaEventType metaEventType, int metaDataLength, long absoluteT
         }
 
         /// <summary>
+        /// Creates a deep clone of this MIDI event.
+        /// </summary>
+        public override MidiEvent Clone() => new MetaEvent(metaEvent, metaDataLength, AbsoluteTime) { data = (byte[])data?.Clone() };
```
jnm2: For example, this deep clone needs to be implemented manually in case `data` is used. However, `data` is not used by any subclass and it's not accessible either, so every subclass will have to override `Clone()` in order to duplicate the original implementation in `MidiEvent.Clone()`.
A solution here might be moving the `data` field to the `UnknownMetaEvent`, since it's null for all other meta events. That way `Clone()` wouldn't have to be overridden in all other `MetaEvent` subclasses.

markheath:
thanks for this, looks good. Your ideas about `UnknownMetaEvent` sound good. I need to think about whether making `MidiEvent` abstract might end up being a breaking change and if so it should be saved for a new major version (1.8) which we are long overdue anyway. Let me know if you want to make those extra changes first, or I can merge as is.

jnm2:
Everybody points me to https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-change-rules.md which says it would be a breaking change because of the public and protected constructors.

I don't really have a strong opinion, but let's not miss making `MidiEvent` abstract in 1.8 if that's something you want to do.

markheath:
OK, I think we should say that next version of NAudio will be 1.8. Its been long enough. So make the abstract change and I'll accept the PR.

jnm2:
Once thing that I really wish I could change is that there are constructors that don't take all the parameters needed to fully initialize. For example, the `SMPTEOffsetMetaMidiEvent` constructor. In order to fully initialize it you have to set properties afterwards.

I'd like to remove the parameterless `MidiEvent` constructor and convert the constructors that take `BinaryReader` into static functions named `Import` for symmetry with the `Export` methods. Then the absolute time and channel would always be passed into `Import` which would do the reading call the constructor which would call the base MidiEvent constructor with the absolute time and channel.

I feel like I could tighten up the API consistency. Are you comfortable with that kind of change?

markheath:
hi @jnm2. These are all good suggestions. I tend to take a fairly conservative approach to breaking changes. If possible I introduce new methods and put the `Obsolete` attribute on the old ones, as a warning that it will go away in vNext, and add a comment saying what should be used instead. Would that work for you if you made a new `SMPTEOffsetMetaMidiEvent` ctor and marked the old one `Obsolete`? Same with the `BinaryReader` constructors - obsolete them and add the new ones. Finally the parameterless `MidiEvent` is protected and looks like it could be removed without likely causing any issues for consumers of the library. 

The MIDI part of NAudio is some of the oldest code in the library so certainly will benefit from some cleanup. I do actually have a number of open source projects that build on it (e.g. MIDI File Mapper and MIDI File Splitter) as well as some closed source apps too. So the obsoleting approach will help for a smoother migration of those projects.

jnm2:
Sure, that's a good idea. Also EditorBrowsable.Never for intellisense?

markheath:
that's not one I was aware of, but sure, would make sense.

jnm2:
We have a bit of a clash. If I leave the obsoleted `ChannelAfterTouchEvent(BinaryReader)` constructor, I will also have to leave and obsolete the `MidiEvent()` constructor. Since `ChannelAfterTouchEvent` ctor calls an obsolete ctor, it generates a compiler warning. You have warnings as errors turned on which is good but will also require use of pragmas every piece of code that touches an obsoleted base ctor.

I think we're going the right direction as far as guaranteeing that the event's invariants are satisfied at construction. I really want to kill that `MidiEvent()` constructor. Another option would be to have `ChannelAfterTouchEvent(BinaryReader)` call `base(default(long), default(int), default(MidiCommandCode))` since that simulates the effect of the parameterless base constructor.

Thoughts?

jnm2:
Also, to avoid duplicating code, I'll have to introduce a helper static method just for the sake of keeping the obsoleted `ChannelAfterTouchEvent(BinaryReader)` constructor:

```c#
public ChannelAfterTouchEvent(BinaryReader br)
    : this(default(long), default(int), ReadAftertouchPressure(br))
{
}

private static int ReadAftertouchPressure(BinaryReader br)
{
    var afterTouchPressure = br.ReadByte();
    if ((afterTouchPressure & 0x80) != 0)
    {
        // TODO: might be a follow-on
        throw new FormatException("Invalid afterTouchPressure");
    }
    return afterTouchPressure;
}

public static ChannelAfterTouchEvent Import(long absoluteTime, int channel, BinaryReader br)
{
    return new ChannelAfterTouchEvent(absoluteTime, channel, ReadAftertouchPressure(br));
}
```

How likely do you think it is that it that the `BinaryReader` constructors are being used from the outside?

markheath:
yeah, thinking about it a bit more and checking through my own library of MIDI projects, I think its actually quite unlikely that anyone is using this since the main way in is through the `MidiFile` class. So let's boot the `BinaryReader` constructors out, and make a clean break with them for 1.8 which I'll probably release soon after we've done this change.

markheath:
just wondering if you're planning to do make any more changes to this PR? I'd like to release NAudio 1.8 fairly soon and for it to include this change

jnm2:
Oh yes, I will try to wrap that up tonight.

jnm2:
My sister's hard drive failed the day before her paper was due and I spent this evening trying to recover it. I'll shoot for tomorrow.

markheath:
no problem - hope you recovered her paper! wasn't meaning to rush you anyway. do it when you have time.

jnm2:
I feel like `MidiEvent` is the wrong abstraction level to be using `NoteLength` and `OffEvent`, but I have to keep the scope from creeping here.

jnm2:
Almost finished, pushed a temporary commit in case you want to critique current progress.

markheath:
only had time for a quick look, but seems good to me. will be nice to get this code cleaned up a bit. If possible can you run the build.bat to make sure all projects are still building.

jnm2:
Yes, it won't build till I'm done with the MetaEvent refactoring which is in progress.

jnm2:
Had a question about "auto sensing"- the MIDI spec says "active sensing." I added an `AutoSensingEvent` and wondered if it should have been an `ActiveSensingEvent`. `MidiEventType.AutoSensing` would have to change or be aliased.

markheath:
hmm, looks like a long-standing typo. Add `ActiveSensingEvent` in there but leave the old one aliased and `[Obsolete]` just in case for backwards compatibility if possible

jnm2:
Another question along the same lines. The spec always has "aftertouch" as a single word but it was pascal-cased as `AfterTouch` in NAudio which implies two words. What do you think about changing it? There is an existing class name `ChannelAfterTouchEvent` and two enum elements.

markheath:
that one's a bit more painful as it will be another breaking change, but we might as well do it and I'll just need to make sure to mention it in the 1.8 release notes.

jnm2:
I figure getting all the pain in at once is best.

jnm2:
I figure getting all the pain in at once is best. These things are however negotiable if you want some and not others.

jnm2:
Noticed we have `MetaEventType.SetTempo` paired with `TempoEvent`, might be fine.

jnm2:
`TempoEvent.Tempo` could be `BeatsPerMinute`

markheath:
probably leave Tempo as is. more people likely to be using that. Hopefully the XML docs mention BPM so should be in the intellisense

jnm2:
It passes tests, the only thing left is for me to figure out which obsolete method signatures should be added. Things like `new MetaEvent` will never be able to work of course since that's now abstract and you're forced to give every event the most derived type.
Maybe you could use this against your projects and paste your error list so I can see what that's going to look like?

jnm2:
I'm sorry this is so much to look through. I rushed to get it all in last night. After we make sure the changes will be compatible, I'll rebase on the latest from master.

Besides the `ActiveSensing` and `Aftertouch` renames, I also removed `MidiEvent.DeltaTime` because that's purely a serialization concern and it wasn't needed, and there's no sensible way to use it when changing the containing list. I also changed deserialization so that the native MIDI command code is preserved such that events coded NoteOn with velocity 0 now come through as NoteOnEvents instead of NoteOffEvents. Do you see a problem with those things?

markheath:
OK thanks for this. I will need to take a good look and think about whether the NoteOnEvent change will be a breaking change for my apps which do a lot of manipulation of note offs. Also need to check MidiFile is not broken by the changes to Absolute time - can't see how it works for type 1 files and if previous event is null. Hopefully will get a chance to try it properly later this week

jnm2:
Good point about MidiFile. I should write a basic test for it.

jnm2:
I just got back from vacation. I think this takes care of type 1 files now. Is there anything else outstanding?

markheath:
thanks, will take another look when I can get some time. My concern with type 1 files was to do with whether `AbsoluteTime` was getting set back to 0 correctly when we start on the events from a subsequent track. Original PR looked like it had lost that behaviour.

jnm2:
You're right, and I have made an edit to that effect and added tests to confirm that is no longer a problem.

markheath:
just a quick update. I am testing this, but there are a lot of breaking changes in my existing MIDI applications especially with things like the parameters for `TempoEvent` swapping round which means the code compiles but has the wrong meaning. Will probably take me a bit longer to test this.

With hindsight would have been better to do this in smaller steps, getting clone in first and a few of the safer refactorings, before bringing in breaking changes as well

jnm2:
@markheath I agree about the smaller steps. The tradeoff was that it made it a lot easier for me and I had been in a hurry to finish it before vacation. I've been worrying about this all along. For the sake of clarity, maybe I should redo the work in steps. It won't be hard because it's still recent in my mind and I have the beginning and end. 

I have a suggestion. I should remove the refactoring commits from this PR so clone can go through, and I'll break up the refactor commit into a few focused PRs.

jnm2:
I should also document every individual API change, such as TempoEvent's constructor, so that 1) you can vet and discuss and 2) we have a nice changelog for the release.
I need to find a good API tool.

jnm2:
What I've done is put the monolithic refactor in a new branch (https://github.com/jnm2/NAudio/tree/midi-refactor-wip) so we can make this PR just about Clone.

Then, I'll start a new issue for discussion of moving to abstract `MidiEvent` and the other items, and what they entail. I intend to document every individual API change, such as TempoEvent's constructor, so that 1) you can vet and discuss and 2) we have a nice changelog for the release.

markheath:
that's great. Would allow us to fast-track all the non-breaking changes in very quickly, and tackle the tricky ones individually

jnm2:
@markheath Let's merge this PR if that's okay so that this is out of the way.
I went through the breaking changes and converted them into an issue for discussion

NAudio/Midi/MidiEvent.cs
markheath: for now, let's get this PR in, and making MidiEvent abstract can be covered by future PR in which this method can be made abstract. Good news is that it is unlikely more derived MidiEvents will actually be created.

NAudio/Midi/MetaEvent.cs
markheath: Have had a look at the code and it would make sense for there to be an `UnknownMetaEvent` (or perhaps `CustomMetaEvent` or `NonStandardMetaEvent`) class. Feel free to introduce one and add that to this PR, as it won't be a breaking change since old code can still create them as `MetaEvent`

markheath:
thanks, have made a couple of comments, am happy to bring this straight in though. Thanks for the thorough writeup of other issues, will try to get round to responding to them all soon

jnm2:
I adapted `RawMetaEvent` which was on the list to add later.
Also, I seem to have a number of commits from your master branch that no longer exist on GitHub. Did you force push or something?

markheath:
that's strange. what are they? I tend to push with Visual Studio GitHub plug-in, and sometimes use the Merge pull request button in GitHub, but I wouldn't expect them to be doing a force.

jnm2:
This is what I got when I fetched upstream/master: https://github.com/jnm2/NAudio/tree/master-missing-commits

Maybe a GitHub glitch?

jnm2:
This is what I got when I fetched upstream/master: https://github.com/jnm2/NAudio/tree/master-missing-commits

Probably somehow I goofed with Visual Studio's rebase onto?

jnm2:
I'll bet I messed something up and you can ignore it.

markheath:
yeah looks like a rebase - same commits, different hashes. This PR merges cleanly so I'll do it now.

PR #118, opened 2016-09-09
Function WriteBigEndian needs an offset of 3

frathihog:
WriteBigEndian was wrong.
If I want to reach the last of 4 bytes,
I must add 3.


markheath:
Cool looks like that was a bug. Have you been able to test this change in any way?

frathihog:
Yes, but it was a short test only.
I tried to manipulate the xing header.

My intention was to fix the length 
for mp3-files with variable length.

frathihog:
Yes, but it was a short test only.
I tried to manipulate the xing header.

My intention was to fix the length 
for mp3-files with variable length.

PR #120, opened 2016-09-28
Update StreamMediaFoundationReader to work with Mono / Unity

bekriebel:
I wanted to be able to use this feature with a unity standalone windows app. I finished adding it to the project with comments and made the changes needed to make it work with Mono's stricter interop requirements. This should continue to work for everything else, but now also works with current versions of Unity if you build it against a .net 3.5 target.

### DIFF
--- a/NAudio/CoreAudioApi/PropVariant.cs
+++ b/NAudio/CoreAudioApi/PropVariant.cs
@@ -55,7 +55,7 @@ namespace NAudio.CoreAudioApi.Interfaces
         [FieldOffset(8)] private short boolVal;
         [FieldOffset(8)] private int scode;
         //CY cyVal;
-        [FieldOffset(8)] private DateTime date;
+        [FieldOffset(8)] private double date;
         [FieldOffset(8)] private System.Runtime.InteropServices.ComTypes.FILETIME filetime;
         //CLSID* puuid;
         //CLIPDATA* pclipdata;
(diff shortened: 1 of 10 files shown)


markheath:
thanks for this. Have you tested that `StreamMediaFoundationReader` still works correctly in the regular windows build? I am also wondering what the significance of setting syncrhonized = false on the ComStream constructor is? Is this something that just applies to Unity/mono? Am happy to bring this in so long as existing behaviour is not impacted

bekriebel:
I did test it against the standard windows .net runtime, though I still only tested with the .net 3.5 framework. I don't see any reason why it would have a problem on the newer versions, though.

As for synchronized=false - yes, this is also a mono required change. Sorry, it snuck into the first commit since it was the first thing I had fixed. From my testing, I didn't notice any difference between it being set to true or false with the standard runtime. If there's any real reason for it, I can look into breaking it off into an option.

BTW, thanks for the library! With it, I now have audio support for cardboard camera photos in my app: https://sprocketvr.com/vrphotoviewer/

markheath:
ok, will try to get this merged fairly soon. The `StreamMediaFoundationReader` was contributed by @softlion so maybe he can check as well that it still works as expected.

markheath:
unfortunately the MethodImpl calls don't compile in the NAudio.Universal.sln and NAudio.Win8.sln solutions. We'd need to surround them with `#if !NETFX_CORE` or similar to make all the solutions compile

bekriebel:
Hrmm... Okay, I'll try to take a look at it sometime in the next few days to see if I can make those projects play nice with the changes.

PR #122, opened 2016-09-29
Add WriteToStream method.

axynos:
Add a method that outputs all data from a WaveProvider to a Stream.
Based off of CreateWaveFile, came in useful for me and thought it would be nice to include in NAudio.

### DIFF
--- a/NAudio/Wave/WaveOutputs/WaveFileWriter.cs
+++ b/NAudio/Wave/WaveOutputs/WaveFileWriter.cs
@@ -54,7 +54,33 @@ namespace NAudio.Wave
                 }
             }
         }
+        
+        /// <summary>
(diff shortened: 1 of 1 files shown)


markheath:
nice idea, I wonder if it might be good to name it more similarly to `CreateWaveFile` maybe `CreateWaveStream` or something more descriptive like `WriteWavFileToStream`

axynos:
Well the current method just appends the data to the end, but doesn't create a new stream, so WriteWavFileToStream would probably be more accurate.

markheath:
ok let's go with that then

markheath:
did you mean to close this PR? I can merge it now if you're done

axynos:
closed by accident, whoops :) And yes I'm done.

PR #124, opened 2016-10-02
AudioFileReader NullReferenceException fix in Dispose

stanhuff:
Here is a fix to an extremely trivial oversight where AudioFileReader wasn't setting its internal stream to null after disposing it, causing further disposes to throw.  Included is a simple integration test using sample wav data provided in the repo.


markheath:
thanks

PR #132, opened 2016-11-06
Reduces Wave32to16Stream allocations

azEf:
The Read method from Wave32to16Stream was allocating a new large array at each call, which caused constant garbage collections.


markheath:
good spot. There is actually a helper in NAudio called `BufferHelpers.Ensure` which reduces the amount of code to do this. It's used like this

     sourceBuffer = BufferHelpers.Ensure(sourceBuffer, numBytes*2);

azEf:
Nice, I made the change.

markheath:
thanks

PR #141, opened 2016-12-05
use correct waveformat for stereo to mono float conversion

phil-n:
Hi @markheath,

first, thanks for this great project.
I tried to convert some .mp3 files from stereo to mono and got an exception (SampleToWaveProvider ctor "Must be already floating point").
The waveformat construction in StereoToMonoSampleProvider looked a bit suspicious, after fixing that it works for me. Since I'm by no means an audio expert, I'd like to have a second pair of eyes check this before merging.

Kind regards Phil

### DIFF
--- a/NAudio/Wave/SampleProviders/StereoToMonoSampleProvider.cs
+++ b/NAudio/Wave/SampleProviders/StereoToMonoSampleProvider.cs
@@ -23,7 +23,13 @@ namespace NAudio.Wave.SampleProviders
                 throw new ArgumentException("Source must be stereo");
             }
             this.sourceProvider = sourceProvider;
-            WaveFormat = new WaveFormat(sourceProvider.WaveFormat.SampleRate, 1);
+            WaveFormat = WaveFormat.CreateCustomFormat(
+                sourceProvider.WaveFormat.Encoding,
+                sourceProvider.WaveFormat.SampleRate,
+                1,
+                sourceProvider.WaveFormat.AverageBytesPerSecond/2,
+                sourceProvider.WaveFormat.BitsPerSample/8,
+                sourceProvider.WaveFormat.BitsPerSample);
         }
 
         /// <summary>


markheath:
hmm, something very weird going on if the original doesn't work for you. What was the exact `WaveFormat` of the incoming `ISampleProvider`?

phil-n:
I'll check when I'm back at work on wednesday. All I remember off the top of my head, encoding was IeeeFloat, 2 channels.

I was using `WasapiOut` with `AudioFileReader` and converted directly via `.ToMono()` extension from there.

markheath:
OK, if you can provide a simple repro that would be great. Sounds like there is a bug somewhere

phil-n:
Alright, back to work earlier than I expected ;).

I can't provide you with one of the original mp3s, since these are voice files from a telephone system, but [here's](https://1drv.ms/u/s!AguVr13SaF8omyGC55fyQsB6ToV3) a small test file without any voice recorded. All of these voice files have the same waveformat:
```
Encoding: IeeeFloat 
Channels: 2 
SampleRate: 8000
AverageBytesPerSecond: 64000 
BlockAlign: 8 
BitsPerSample: 32 
ExtraSize: 0
```

I'm using this code to play the file (straight from LinqPad):
```csharp
void Main()
{
    var output = new WasapiOut(AudioClientShareMode.Shared, 50);
    var provider = new AudioFileReader(@"D:\testfile.mp3");
    provider.WaveFormat.Dump();
    output.Init(provider.ToMono());

    var play = new Button
    {
        Content = "Play",
        Width = 120,
        Height = 28
    }.Dump();
    play.Click += (s, e) => output.Play();

    var stop = new Button
    {
        Content = "Stop",
        Width = 120,
        Height = 28
    }.Dump();
    stop.Click += (s, e) =>
    {
        output.Stop();
        provider.Position = 0;
    };
}
```

which then aborts with the following exception:
```
ArgumentException "Must be already floating point"

   at NAudio.Wave.SampleProviders.SampleToWaveProvider..ctor(ISampleProvider source)
   at NAudio.Wave.WaveExtensionMethods.Init(IWavePlayer wavePlayer, ISampleProvider sampleProvider, Boolean convertTo16Bit)
   at UserQuery.Main()
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.ThreadHelper.ThreadStart()
```

I've also tested this with some generic music mp3 with the following waveformat and didn't get that exception:
```
Encoding: IeeeFloat 
Channels: 2 
SampleRate: 44100 
AverageBytesPerSecond: 352800 
BlockAlign: 8 
BitsPerSample: 32 
ExtraSize: 0 
```

markheath:
thanks, have tracked it down. There is a helper method on `WaveFormat` for this purpose, so I used that instead.

PR #142, opened 2016-12-08
[WasapiCaptureRT] Init method; StartRecording waits for actual stop

kamenlitchev:
Several changes for WasapiCaptureRT:

### Introduce InitAsync() method to explicitly initialize WASAPI from the UI thread
Consequently, the client can call StartRecording from any thread they want. In fact, StartRecording can be called even now from any thread, but a separate InitAsync method makes the UI-thread requirement a more prominent one.
InitAsync() will be called internally by StartRecording() if not called before, so that we do not break existing implementations

### StopRecroding() will now properly wait for the actual WASAPI recording to stop 
One TO-DO less

### AudioClient resources will be cleaned in the Dispose method
...and not in the DoRecording method. Not sure if the original idea was not that the Start/Stop cycle be actually a Create/Dispose cycle.


kamenlitchev:
The need to separate the UI-thread exclusive Init calls was that in our use-case scenario, the start of the recording is handled in an external library (which uses NAudio) in a background thread due to blocking operations. Thus, we need a lightweight "Ask Permissions" Init call without actually starting the WASAPI capture.

markheath:
Can't get this to compile. Am I missing something (building the NAudio.Universal.sln)?

```
WasapiCaptureRT.cs(299,30,299,42): error CS7036: There is no argument given that corresponds to the required formal parameter 'callback' of 'EventHandler<StoppedEventArgs>.BeginInvoke(object, StoppedEventArgs, AsyncCallback, object)'
```

kamenlitchev:
Sorry - was experimenting with simplifying the Task.Run line for the event with something simpler. Committed the wrong code.

kamenlitchev:
I changed RaiseRecordingStopped to match WasapiOutRT.RaisePlaybackStopped

PR #151, opened 2017-01-06
Add StreamMediaFoundationReader support for windows 7

joewen:
When trying to use StreamMediaFoundationReader on windows 7. A exception will occur due to MFCreateMFByteStreamOnStreamEx(https://msdn.microsoft.com/en-us/library/windows/desktop/hh162754(v=vs.85).aspx) support only windows 8 and above. So I added another MediaFoundation api MFCreateMFByteStreamOnStream(https://msdn.microsoft.com/zh-tw/library/windows/desktop/dd388095(v=vs.85).aspx) into MediaFoundationInterop and use define in MediaFoundationApi.CreateByteStream to seperate RT and desktop apps Implemention.

### DIFF
--- a/NAudio/MediaFoundation/MediaFoundationHelpers.cs
+++ b/NAudio/MediaFoundation/MediaFoundationHelpers.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Runtime.InteropServices;
 using System.Text;
 using NAudio.Wave;
+using System.Runtime.InteropServices.ComTypes;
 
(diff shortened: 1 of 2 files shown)


markheath:
thanks for this - looks good from code review and compiles, so I've merged it in. What operating systems were you able to test on?

joe791023:
Tested on Windows 7 and Windows 10. Both works great!

joewen:
Tested on windows 7 and 10. Both works great!

PR #159, opened 2017-01-28
expose AsioOut.FramesPerBuffer

YaakovDavis:
Note: I use the term "Frames", as it's more precise than "Samples". (The latter might refer to the total, or per-channel count. See the comment in code).


markheath:
great, thanks for this

PR #172, opened 2017-03-03
Change WaveOut and WaveOutEvent default DeviceNumber to -1

joewen:
Setting DeviceNumber to 0 means current default device at that time. The sound will come out from same device if the default device changed while playing. If we set DeviceNumber to -1, the output device will always stick to the default device automatically.

Although this wasen't mention in waveOutOpen [document](https://msdn.microsoft.com/zh-tw/library/windows/desktop/dd743866(v=vs.85).aspx). Some other projects had already set DeviceNumber to -1. https://www.codeproject.com/Articles/866347/Streaming-Audio-to-the-WaveOut-Device 
`waveOutOpen(out hwo, -1, wFmt, woDone, rFrames, CALLBACK_FUNCTION);`

### DIFF
--- a/NAudio/Wave/WaveOutputs/WaveOut.cs
+++ b/NAudio/Wave/WaveOutputs/WaveOut.cs
@@ -66,9 +66,10 @@ namespace NAudio.Wave
         /// <summary>
         /// Gets or sets the device number
         /// Should be set before a call to Init
-        /// This must be between 0 and <see>DeviceCount</see> - 1.
+        /// This must be between -1 and <see>DeviceCount</see> - 1.
+        /// -1 means stick to default device even default device is changed
         /// </summary>
-        public int DeviceNumber { get; set; }
+        public int DeviceNumber { get; set; } = -1;
 
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


markheath:
what operating systems have you tried this on? NAudio currently supports back to Windows XP, and I'd like to avoid breaking any existing behaviour.

joewen:
I tested on Windows 7 and 10. This will change existing behavior when using default DeviceNumber. Or maybe I should remove code changes then keep only comment added in this pr is better?

markheath:
Looks like this is basically WAVE_MAPPER and seems to have been around for a while. I'll give it a quick run-through myself when I get a chance and then I'll merge this in

markheath:
looks good. I'll revert if anyone reports any issues with this, but seems a sensible default

MagicMau:
This does not actually change the default it seems, as the constructor of WaveOut and WaveOutEvent explicitly sets DeviceNumber = 0. So you would still have to set it manually to -1.

markheath:
Good spot. I've updated the code and given it a quick test.

PR #180, opened 2017-04-15
Added MidiFile constructor overload that takes an input Stream object.

object:
Added a MidiFile constructor overload to be able to read MIDI file events from an input stream.
The implementation is similar to those of AiffFileReader and Mp3FileReader with one difference. Unlike AiffFileReader or Mp3FileReader, MidiFile doesn't implement IDisposable so I couldn't dispose the stream in Dispose method and didn't want to introduce dependency on IDisposable for the sake of backward compatibility. So I added a private constructor with ownInputStream parameter which is checked in "finally" clause - I had to replace "using" clause with "try/finally" because of that.

### DIFF
--- a/NAudio/Midi/MidiFile.cs
+++ b/NAudio/Midi/MidiFile.cs
@@ -39,12 +39,27 @@ namespace NAudio.Midi
         /// </summary>
         /// <param name="filename">Name of MIDI file</param>
         /// <param name="strictChecking">If true will error on non-paired note events</param>
-        public MidiFile(string filename, bool strictChecking)
+        public MidiFile(string filename, bool strictChecking) :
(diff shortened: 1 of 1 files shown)


markheath:
thanks for this. Have you checked that all the NAudio solutions still build with this change?

object:
I rebuilt only the main solution, will try others asap.

object:
Yes, NAudio.Universal solution builds too. I don't have necessary components installed to build Win8 solution but I am quite certain it builds since all others build OK and I just provided an overload similar to what was already present in a few other classes.

markheath:
... I've not forgotten this - just been a crazy busy few weeks. Hope to merge this next week :)

object:
No problem! Updated NuGet package will also be much appreciated :-)

Issue #200, opened 2017-06-08
Memory Leak in AudioEndpointVolume


alexunderattack:
Hello, we have an implementation of sound controls with the typical actions (Mute, slider volume management, sample audio input / output, etc.), and we have a significant memory leak. The application is built by attacking the audio dll in a decoupled way through an intermediate wrapper through interfaces to be able to do unit tests and mock the system audio cards.

The problem is that any access to AudioEndpointVolume to query or modify the volume, constantly creates us COM objects that are not destroyed with Dispose.

The management of the MMDevices objects is correct, but the AudioEndpointVolume and are a property of the MMDevice. Should we make an interface inside the wrapper that also implements the Interface IAudioEndpointVolume and make inside the wrapper our deletion of the COM objects it creates or is there a simpler way to make a dispose of the AudioEndPointVolume COM objects?

<a href="https://ibb.co/hHHT5v"><img src="https://preview.ibb.co/iBJCXa/Captura.png" alt="Captura" border="0"></a><br /><a target='_blank' href='https://es.imgbb.com/'>fotos imagenes</a><br />

alexunderattack:
Hi,
Finally and since the impact of implementing the internal interface of IAudioEndpointVolume is very large, we have chosen to explicitly destroy all the COM objects of that interface through the GUID of the interface.

 `  try
            {
                //unable to dispose all unmanaged COM objects of IAudioEndpointVolume internal interface
                // of Naudio. The only way is delete objects by the Interface GUID

                System.ComponentModel.GuidConverter gc = new System.ComponentModel.GuidConverter();

                System.Guid guid = (Guid)gc.ConvertFromString(GUID_NAUDIO_COMOBJECTS);

                Type consoleType = Type.GetTypeFromCLSID(guid);

                object objConsole = Activator.CreateInstance(consoleType);

                System.Runtime.InteropServices.Marshal.ReleaseComObject(objConsole);

            }
            catch (Exception ex)
            {

                Log.WarnFormat(CultureInfo.InvariantCulture,"Fail release COM object Audio Card ->{0}", ex.Message);

            }`

markheath:
The intersection of .NET's garbage collection with COM's reference counting makes for some headaches working out the best way to track the lifetimes of these objects. It's an area of NAudio I'd like to revisit at some point.

PR #212, opened 2017-07-24
Implemented IDisposable into MMDevice and added releasing of COM object to AudioEndpointVolume Dispose

simon-wh:
Should help with #200 as the unused AudioEndpointVolume COM objects can be easily removed if the MMDevice instance is no longer being used to prevent memory leaks.


markheath:
thanks, had to make an additional fix due to some dispose issues with the Volume Mixer demo, but seems OK now

PR #220, opened 2017-08-06
Allow for poorly formatted Chunk Identifiers in WAV header by seeking…

ArclightSA:
… near expected identifier position.

I ran into a problem where a large number of WAV files were created by a professional editor, which appended an "inst" chunk with a length of 7 bytes but the following "data" chunk was 8 bytes away.
WaveFileChunkReader threw a FormatException because it was unable to locate the data chunk.
It seems that this type of thing can happen.
https://sites.google.com/site/musicgapi/technical-documents/wav-file-format#formatvariations

This fix just seeks locally to see if there is a one byte variance in chunk length.

I doubt this problem affects many people otherwise it'd have been handled already.

### DIFF
--- a/AudioFileInspector/Properties/AssemblyInfo.cs
+++ b/AudioFileInspector/Properties/AssemblyInfo.cs
@@ -29,8 +29,8 @@ using System.Runtime.InteropServices;
 //      Build Number
 //      Revision
 //
-[assembly: AssemblyVersion("0.1.7.0")]
-[assembly: AssemblyFileVersion("0.1.7.0")]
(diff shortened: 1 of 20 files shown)


markheath:
thanks for this contribution. it's on hold at the moment until I have some time to test it. I need to understand what the performance implications are, and whether it could cause corrupted sound to be played with a malformed WAV file

markheath:
Am closing this as NAudio 3 codebase has diverged significantly. Thanks for the contribution, and feel free to raise an issue if you have an example of a WAV file that NAudio can't play

PR #224, opened 2017-08-08
Stop with DirectSound, prevent noise from playback

Hondarer:
With Directsound, when another playback is started after the playback has completely stopped, the problem of mixing noise is solved.

----

In DirectSound, when playback is started,
the rest of the sound that was played last time is played back as noise.
This happens even if the secondary buffer is completely silenced, 
so it seems that the buffer in the primary buffer or higher is not cleared.

To solve this problem fill the secondary buffer with silence data when stop playback.

### DIFF
--- a/NAudio/Wave/WaveOutputs/DirectSoundOut.cs
+++ b/NAudio/Wave/WaveOutputs/DirectSoundOut.cs
@@ -550,40 +550,7 @@ namespace NAudio.Wave
             {
                 if (secondaryBuffer != null)
                 {
-
-                    // **************************************************
-                    // In DirectSound, when playback is started,
-                    // the rest of the sound that was played last time is played back as noise.
-                    // This happens even if the secondary buffer is completely silenced, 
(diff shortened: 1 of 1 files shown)


NAudio/Wave/WaveOutputs/DirectSoundOut.cs
```diff
@@ -550,6 +550,41 @@ private void StopPlayback()
             {
                 if (secondaryBuffer != null)
```
markheath: might be nice to factor this whole if statement into a helper method - e.g. CleanUpSecondaryBuffer

Hondarer: Thank you for reviewing.
Based on the advice of @markheath , I cleared the secondary buffer clearing process as a helper method at c74756b0c2aa86066a579c8c86f5d442dc154a69 .

markheath:
thanks, have been on holiday but will try to get this tested and merged soon

PR #230, opened 2017-09-02
MMDevice property manipulation functionality for NAudio

Bannar:
Key changes:
- PropVariant fields now publicly alterable
- Updates PropertyKeys dictionary with new PropertyKey static fields
- MMDevice::GetPropertyInformation method now public and allows for access mode change
- Implemented SetValue and Commit methods for PropertyStore class.

Changes listed above are targeted to allow user to modify MMDevice properties dynamically e.g: endpoint's displayed name.

Example code:
```
PropertyKey propKey = PropertyKeys.PKEY_Device_DeviceDesc;
IntPtr pointer = Marshal.StringToHGlobalAuto("hello world");
PropVariant value = new PropVariant()
{
    vt = (short)VarEnum.VT_LPWSTR,
    pointerValue = pointer
};

// deviceId string, enumerator MMDeviceEnumerator
MMDevice device = enumerator.GetDevice(deviceId);

// by default, read-only flag is used which causes Access Denied exception to occur
device.GetPropertyInformation(StorageAccessMode.ReadWrite);

PropertyStore store = device.Properties;
store.SetValue(propKey, value);
store.Commit();
// free memory
Marshal.FreeHGlobal(pointer);
```

### DIFF
--- a/NAudio/CoreAudioApi/Interfaces/Blob.cs
+++ b/NAudio/CoreAudioApi/Interfaces/Blob.cs
@@ -27,9 +27,18 @@ using System.Runtime.InteropServices;
 
 namespace NAudio.CoreAudioApi.Interfaces
 {
+    /// <summary>
+    /// Representation of binary large object container.
+    /// </summary>
     public struct Blob
     {
+        /// <summary>
+        /// Length of binary object.
+        /// </summary>
         public int Length;
+        /// <summary>
+        /// Pointer to buffer storing data.
+        /// </summary>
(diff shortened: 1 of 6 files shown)


markheath:
thanks for this contribution - seems like it will be useful. It doesn't compile due to missing XML documentation for new public properties. In NAudio, all public types and members must have XML docs. If you can fix this, I will get it tested

Bannar:
Appended new XML documentation for public members as asked.

PR #233, opened 2017-09-04
Improved support for mono AAC

Ahmed-Abdelhameed:
Multiplying the SampleRate & ChannelCount will cause exception with aac_lc mono _(it's only required with HE-AAC)_.
This solution was suggested by @tfwio. It's a dirty fix, _but it works_. I'll be trying to find a better solution as well (an ideal one would be to differentiate the LC format profile form HE for the AAC audio format).

### DIFF
--- a/NAudio/Wave/WaveStreams/MediaFoundationReader.cs
+++ b/NAudio/Wave/WaveStreams/MediaFoundationReader.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Diagnostics;
 using System.IO;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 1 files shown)


markheath:
ok shame as this is a bit ugly. we should catch the most specific exception possible - Exception type and even a when clause if possible
Also, do you have the capability to test this fix on both Win 7 and Win 10, since the behaviour seems to be different?

Ahmed-Abdelhameed:
Yes, I know. I'm also looking into MS documentation trying to find a way to differentiate between the two format profiles.
Anyhow, will test on Win10 right now, and come back with the result.

Ahmed-Abdelhameed:
Hmm, actually there **is** a difference in the behavior..

 Unlike Win7, on Win10, passing the `IMFMediaType` object with wrong SampleRate/ChannelCount does **Not** cause an exception to be thrown _(apparently, that's what caused [the original issue](https://github.com/Ahmed-Abdelhameed/NAudio/commit/a25f271b3802cdb4859878489286667d8cc214d3))_.

So unfortunately, applying this fix, will only make things work fine for Windows 7, but _some audio files_ won't be read correctly on Win10 (back to the original issue again) :/

If only we could differentiate between the two profiles (LC & HE/v2), I believe everything will work perfectly.
I tried to find something useful in [the documentations](https://msdn.microsoft.com/en-us/library/windows/desktop/dd742785(v=vs.85).aspx) with no luck. It might have something to do with [`MF_MT_AAC_AUDIO_PROFILE_LEVEL_INDICATION`](https://msdn.microsoft.com/en-us/library/windows/desktop/dd319560(v=vs.85).aspx) attribute, or [`MF_MT_USER_DATA`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms693533%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396), but I couldn't get anything to work so far.

markheath:
ok, thanks for looking into this. hopefully you can find a solution. We can even probe the version of Windows if necessary. You might find someone who can help on the [MSDN forums](https://social.msdn.microsoft.com/Forums/en-US/home?forum=mediafoundationdevelopment). 

It would also be a good idea to collect some example audio files we can use for integration tests. I could put them in an Azure blob container. Would need to not contain copyrighted material though.

Ahmed-Abdelhameed:
I couldn't found something helpful on MDSN _yet_. However, I made some improvements to when exactly to change the sample rate & channels, and made the `catch`ing more accurate.

Please check the last change. This was tested on both Windows 7 & 10, it works perfectly fine. The following audio formats were used for testing:

- HE-AAC - in an MP4 container.
- AAC-LC mono & stereo.
- AAC-LC in an M4A container.
- AAC-LC mono with low sample rate (8000).

markheath:
cool, I will do some additional testing of my own, but looks good

Ahmed-Abdelhameed:
Yeah, of course. Also, kindly update the NuGet package when possible.
Thank you for your quick support :)

markheath:
new NuGet 1.8.3 pushed.

PR #236, opened 2017-09-05
Prevent audio files from staying locked by MediaFoundationReader class

Ahmed-Abdelhameed:
Fixing an issue that causes audio files used by the MediaFoundationReader class to stay locked after the class instance is disposed, due to unreleased COM object.

### DIFF
--- a/NAudio/Wave/WaveStreams/MediaFoundationReader.cs
+++ b/NAudio/Wave/WaveStreams/MediaFoundationReader.cs
@@ -101,6 +101,10 @@ namespace NAudio.Wave
             {
                 pReader = reader;
             }
+            else
+            {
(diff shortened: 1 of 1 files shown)


mattjamer6962:
Mark, the latest version of the naudio nuget package v 1.8.3 doesn't include the changes you made to fix this issue on Sep 7 as it was released on Sep 5. I was encountering a similar file locking issue and once I pulled the latest code from github, compiled it, and used that instead of the nuget package it worked as expected with no file locking issues. Can you please rebuild and publish an updated version of the nuget package?

markheath:
NAudio v1.8.4 now released with this fix

mattjamer6962:
Great! Thanks Mark.

PR #257, opened 2017-11-03
Update SilenceWaveProvider.cs

MagicMau:
Hi,

I am using the SilenceProvider to keep my mixer running and noticed it's using a for-loop to zero out the buffer. I replaced it with Array.Clear() as it is much faster (https://msdn.microsoft.com/en-us/library/system.array.clear%28v=vs.110%29.aspx). 
Might be of use to others as well?


markheath:
thanks for this. there are a few nasty gotchas around using `Array.Clear()` with the `WaveBuffer` class but in this case we are safe as it is operating on a `byte[]` not `float[]`.

PR #260, opened 2017-11-09
Fixed issue reading a MP3 file that does not have a ID3v2 tag.

lmh9974a:
Hello,

I encountered an error when reading a MP3 file. The error was "Not an ID3v2 tag".

The MP3 files did not have a ID3v2 tag, just a ID3v1 tag. I removed the exception. The lib now reads the file without any problem.

### DIFF
--- a/NAudio/FileFormats/Mp3/Id3v2Tag.cs
+++ b/NAudio/FileFormats/Mp3/Id3v2Tag.cs
@@ -212,15 +212,16 @@ namespace NAudio.Wave
                     // footer present
                     byte[] footer = reader.ReadBytes(10);
                 }
+
+                tagEndPosition = input.Position;
+                input.Position = tagStartPosition;
+                rawData = reader.ReadBytes((int)(tagEndPosition - tagStartPosition));
             }
             else
             {
                 input.Position = tagStartPosition;
-                throw new FormatException("Not an ID3v2 tag");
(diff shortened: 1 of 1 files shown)


markheath:
thanks for this contribution. I will need to test it against more MP3 files to ensure no regressions before merging this in. NAudio can already read files without ID3v2 tags, this seems to be a corrupt ID3v2 tag issue, or perhaps a false positive

EveryOtherUsernameWasAlreadyTaken (commented):
If the Id3v2Tag constructor does no longer throw an exception, ReadTag() might no longer need to catch it.

xPaw:
> NAudio can already read files without ID3v2 tag

I don't see how, constructing `Mp3FileReaderBase` does `Id3v2Tag.ReadTag` as the first thing on the stream, which throws as it does not see the tag.

xPaw:
The thrown exception is caught here: 
https://github.com/naudio/NAudio/blob/a106da4eed61774e9bd3eda1fa7922581aee04e1/NAudio.Core/FileFormats/Mp3/Id3v2Tag.cs#L21-L31

I can confirm that mp3 files without ID3v2 tag do play (just that Visual Studio was catching the exception).

markheath:
closing this as it has been resolved a different way in main (`Id3v2Tag.Read`). Thanks for your contribution

PR #261, opened 2017-11-14
WaveIn object no longer stops recording

Protocat112:
Fixed issue with WaveIn object setting recording to false on its own
when mutiple WaveIn objects are used, and after recording has been
stopped and then started again.

I don't know if this is the best way to do it, but it fixed the bug for my program, which uses multiple WaveIn objects for multiple streams. If I recreated the WaveIn objects (to allow for a clean change of input device for that object), and tried to restart recording, exactly every other time, the 'recording' property would be set to false, due to an MmException.

### DIFF
--- a/NAudio/Wave/WaveInputs/WaveIn.cs
+++ b/NAudio/Wave/WaveInputs/WaveIn.cs
@@ -142,10 +142,9 @@ namespace NAudio.Wave
                     {
                         buffer.Reuse();
                     }
-                    catch (Exception e)
+                    catch (MmException)
                     {
-                        recording = false;
-                        RaiseRecordingStopped(e);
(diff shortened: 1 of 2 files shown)


Protocat112:
Thanks for the help, I will look into that!

Protocat112:
WaveInterop.waveInOpenWindow() is now giving me an UnspecifiedError, but only on the last channel I try, which is strange, any idea why this may be happening? Also, looking at your documentation, I am creating a GUI application not a console app, but the form is running on a separate thread. Is WaveInEvent still the best thing to call?

Protocat112:
I have decided to just catch the exception and move on, and everything seems to work now! I think using WaveInEvent helped somewhat :)

PR #312, opened 2018-03-09
StereoToMonoSampleProvider.Read fills incorrect range

azyobuzin:
I found that `StereoToMonoSampleProvider.Read` would not work when I specify non-zero value to `offset` parameter. The method fills the range which starts from half of the `offset` I specified with samples.

I fixed it and wrote a test which tests whether the method fills the specified range and not fills the rest.


markheath:
many thanks

PR #313, opened 2018-03-09
Test "CanReuseWasapiCapture" never ends

azyobuzin:
# Problem
I found that the test case `CanReuseWasapiCapture` never ends when I run all of the tests.

The cause of this issue is that it calls `Dispose` via `using` statement immediately after calling `StartRecoding`.

`Dispose` calls `StopRecoding`, and the method sets `Stopping` to `captureState`. `DoRecoding` on the new thread made in `StartRecoding` will set `Capturing` to `captureState` and start recoding even if `captureState` is `Stopping`. So when the new thread is started after `StopRecoding` is called, the recoding will not stop until we call `StopRecoding` again. As to `Dispose`, the method never returns because the method waits for the recoding thread.

# Solution
I made `CaptureThread` and `DoRecoding` check whether `captureState` is the expected value using `Interlocked.CompareExchange` (The type of `captureState` is changed to `int` because of using it). If the value is not the expected value, it considers the situation as when `StopRecoding` is called before the thread starts and does termination.

# Test
I ensured that there is no new failed test case and WASAPI Capture in WPF Demo works.

### DIFF
--- a/NAudio/Wave/WaveInputs/WasapiCapture.cs
+++ b/NAudio/Wave/WaveInputs/WasapiCapture.cs
@@ -38,7 +38,7 @@ namespace NAudio.CoreAudioApi
     {
         private const long ReftimesPerSec = 10000000;
         private const long ReftimesPerMillisec = 10000;
-        private volatile CaptureState captureState;
+        private volatile int captureState;
         private byte[] recordBuffer;
         private Thread captureThread;
         private AudioClient audioClient;
@@ -114,7 +114,11 @@ namespace NAudio.CoreAudioApi
         /// <summary>
         /// Current Capturing State
         /// </summary>
-        public CaptureState CaptureState {  get { return captureState; } }
+        public CaptureState CaptureState
+        {
+            get { return (CaptureState)captureState; }
+            set { captureState = (int)value; }
+        }
 
         /// <summary>
         /// Capturing wave format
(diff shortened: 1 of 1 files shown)


NAudio/Wave/WaveInputs/WasapiCapture.cs
```diff
@@ -224,28 +228,54 @@ public void StartRecording()
         /// </summary>
         public void StopRecording()
         {
-            if (captureState != CaptureState.Stopped)
-                captureState = CaptureState.Stopping;
+            // Set Stopping to CaptureState if the value is not Stopped
+            while (true)
```
markheath: the idea behind the stop recording method is that it only requests that recording stops. It does not block until stopping has complete. This change looks like it will max out CPU until current state gets set to stopped.

markheath:
I think I understand the issue here, but would be better to fix in a different way that doesn't change the behaviour of `StopRecording`. That unit test isn't really following good practices anyway, but the race condition you've hit should be fixed

NAudio/Wave/WaveInputs/WasapiCapture.cs
azyobuzin: It does not wait for stopping. It assigns `Stopping` to `currentState` using `CompareExchange` pattern to prevent from overwriting the field when the recoding thread writes to `currentState` at the same time.

Issue #339, opened 2018-06-04
WaveOutEvent.Dispose() thread safety issues


neilt6:
`WaveOutEvent.Dispose()` makes sure to call `WaveOutEvent.Stop()`, but `WaveOutEvent.Stop()` doesn't wait for the background thread to exit before returning. This is causing intermittent exceptions in my application. I propose `WaveOutEvent` be modified to use a proper `Thread` that can be joined in `Stop()` instead of a fire-and-forget `ThreadPool` thread.

markheath:
Have you got a stack trace of the exception you're getting? Would like to see if I can repro.

neilt6:
Unfortunately no. The nature of the bug makes it very hard to reproduce, and today it's not happening.

markheath:
You can always subscribe to `PlaybackStopped` to know that `Dispose` is safe to call. But closing this old issue as `WasapiPlayer` is now the recommended way to play audio on Windows going forwards in NAudio 3.

PR #340, opened 2018-06-04
Fixed WaveOutEvent.Dispose() race condition

neilt6:
Added code to ensure the playback thread exits before `Stop()` returns to fix a race condition that occurs if `Dispose()` is called on a running player. Fixes issue #339.

### DIFF
--- a/NAudio/Wave/WaveOutputs/WaveOutEvent.cs
+++ b/NAudio/Wave/WaveOutputs/WaveOutEvent.cs
@@ -18,6 +18,7 @@ namespace NAudio.Wave
         private IWaveProvider waveStream;
         private volatile PlaybackState playbackState;
         private AutoResetEvent callbackEvent;
+        private Thread playbackThread;
         private float volume = 1.0f;
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


neilt6:
@markheath Would you mind taking a look at this pull request?

markheath:
`Stop` is not intended to be a blocking operation and it would be a breaking change to make it so. So I'd prefer to find another way to resolve this issue. The recommended pattern to stop playback in NAudio is to request that playback stops, then wait for it to actually stop (`PlaybackStopped` event is raised), and then you are free to `Dispose`. See this document for more info: https://github.com/naudio/NAudio/blob/master/Docs/PlaybackStopped.md

neilt6:
I see your reasoning. In this case I believe `Dispose()` should be made blocking somehow even if `Stop()` is non-blocking. It's not always possible to call `Stop()` and wait for `PlaybackStopped` before calling `Dispose()`.

markheath:
One thing I have wanted for some time is a new "Stopping" (and probably Starting) state on PlaybackState. This would allow the state to be set to "Stopping" when you call `Stop`, and so in the `Dispose` method, we could wait up to a set amount of time for the playback thread to exit. It's something we could possibly get into v1.9

PR #356, opened 2018-07-02
ASIO: Fix sample converter with floats

LAGonauta:
When recording from ASIO and sending the output to another ASIO device I noticed that the input and the output were Int32LSB. This caused problems with NAudio ASIO implementation as the sample converter currently do not support Int32LSB.

I did some modifications to NAudio and now it seems to be working as it should, but I could test only Int32LSB as input and Int32LSB as output.


markheath:
thanks for this. haven't been able to test these conversions myself, but looks good from code review

PR #376, opened 2018-08-01
Update CueWaveFileReader.cs

Vashq8:
Cue read from wave stream, I thought I proposed the same earlier.


markheath:
thanks for this contribution

Yitzchok:
This PR breaks the build.
This should be base(inputStream) instead of this(inputStream).

markheath:
Ah thanks for pointing this out. Hoping to get Azure Devops pipelines integration working soon to detect non-compiling PRs automatically.

PR #396, opened 2018-10-09
Added 32 bits WAV to 8 conversion

MarcosCobena:
Hey,

My mate @vicfergar wrote the algorithm to convert from 32-16 to 8 bits, and since we're consuming NAudio in our project for everything else related, we moved on and translated the former into NAudio, so anyone else can use it too and even we can get rid of the code with future NuGet packages.

Let us know any change you see strange.

Thanks in advance for your work!

### DIFF
--- a/NAudio/NAudio.csproj
+++ b/NAudio/NAudio.csproj
@@ -511,6 +511,7 @@
     <Compile Include="Dmo\InputStreamInfoFlags.cs" />
     <Compile Include="Dmo\OutputStreamInfoFlags.cs" />
     <Compile Include="Dmo\ResamplerMediaObject.cs" />
+    <Compile Include="Wave\WaveProviders\WaveFloatTo8Provider.cs" />
   </ItemGroup>
   <ItemGroup>
     <EmbeddedResource Include="Gui\Fader.resx">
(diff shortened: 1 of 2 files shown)


MarcosCobena:
Let me write down I've been working always with Visual Studio for Mac (hence macOS), so needed to change .NET Framework 3.5 Client Profile requirement to .NET Framework 3.5 —plus some other minor changes in CSPROJs. None of those changes are included in the PR.

I did build "everything" up but didn't run unit tests. My confidence with this PR comes more from [this other repo](https://github.com/MarcosCobena/ManagedAudioLibraries), where we've done all the job, and does have unit tests for such piece.

markheath:
thanks for this contribution. Code submitted to NAudio should follow the existing naming conventions and code formatting.

Also, looks to me like the max sample value with this code would be 254 rather than 255. Was that intentional?

MarcosCobena:
Hi Mark,

Thanks for your time reviewing.

Would you have close any link to the conventions used? I remember checking that, but didn't found any.

Sorry but I don't understand which max sample value you mention, would you mind highlighthing some lines in the code?

For sure, I'll make those changes needed in order to conform the repo :-)

Thanks again.

PR #399, opened 2018-10-11
Added WaveIn.GetPosition and WaveInEvent.GetPosition calling through …

Mostlypyjamas:
This adds in a new GetPosition method to WaveIn and WaveInEvent that calls through to waveInGetPosition, as previously there was no way to retrieve the current number of bytes that have been recorded.

This uses the same pattern as the WaveOut.GetPosition method.


markheath:
great, thanks for this

PR #403, opened 2018-10-16
Added WaveInReset (required for MM to trigger callback). Moved callback Reset trigger

Mostlypyjamas:
Changed WaveInEvent to call WaveInReset after WaveInStop to trigger buffers being returned via the callback as per https://msdn.microsoft.com/en-us/library/windows/desktop/dd743850(v=vs.85).aspx

Also moved callbackEvent.Set to below waveInStop+new waveInReset so those have completed BEFORE the callback look processes those buffers to check for any audio bytes recorded.

### DIFF
--- a/NAudio/Wave/WaveInputs/WaveInEvent.cs
+++ b/NAudio/Wave/WaveInputs/WaveInEvent.cs
@@ -186,8 +186,12 @@ namespace NAudio.Wave
             if (captureState != CaptureState.Stopped)
             {
                 captureState = CaptureState.Stopping;
-                callbackEvent.Set(); // signal the thread to exit
                 MmException.Try(WaveInterop.waveInStop(waveInHandle), "waveInStop");
+
+                //Reset, triggering the buffers to be returned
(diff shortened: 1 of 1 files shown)


markheath:
Thanks for this contribution. I am quite cautious about accepting changes in this area since this is one of the classes that is most commonly used in NAudio and needs to work with a wide variety of soundcard drivers. I'll have to give this a try myself.
Can you tell me a bit about what testing you've performed on this.
Do you definitely need Reset as well as Stop, or does just calling Reset also work?

markheath:
OK, have tested it and seems good. I made one small change to avoid calling `DataAvailable` if the `BytesRecorded` was 0.

Mostlypyjamas:
Hi Mark, apologies for not coming back to you sooner, glad you're happy with it! - As I suspect you may have found in your own testing now, the main reason is that a Stop doesn't trigger the callback whilst the reset does, and whilst you can then poll the buffers to see their state, it's undefined in the MSDN documentation as to whether stop will have marked them as ready for receiving.

Glad I could help though, cheers!

PR #413, opened 2018-10-29
Added DMO Effector and DmoEffectWaveProvider

RestNote:
I added an effector using DMO(DirectX Media Object). 
Also, to make it easier to handle in the actual implementation, we added the Wave Provider using the DMO effector.


markheath:
thanks, this looks really cool. I'll see if I can find some time to try it out first, and then get it merged

PR #436, opened 2019-01-25
Added 24-bit to float and float to 24-bit wave providers.

jobrittain:
Hello,

We needed 24-bit to 32-bit IEEE float conversion so I implemented them for our project. So far they've only been tested with some Audacity converted samples, but they work.


krixgris (approved):

markheath:
Thanks for this contribution. This need is met by `Pcm24BitToSampleProvider` and `SampleToWaveProvider24`  so I don't think we need to add an alternative way to achieve the same thing. Let me know if there is a specific use case that you are not able to implement

PR #438, opened 2019-01-25
Prevent deadlock in WasapiCapture

tombogle:
Used locking to make the WasapiCapture class prevent deadlock when it gets a call to stop recording before it has had a chance to start capturing.


markheath:
Thanks for this contribution, although I will close it as NAudio 3 has been significantly updated in this area (and `WasapiCapture` itself s now replaced with `WasapiRecorder`)

Issue #498, opened 2019-06-25
NAudio.Midi.MidiFile.Export() to support stream


MaKiPL:
Currently NAudio.Midi.MidiFile.Export() contains only one variant:

Export(string, MidiEventCollection)
I'm asking for an overload of this method to support supplied stream instead of limiting the method to file I/O
Example:
Export(System.IO.Stream, MidiEventCollection)
That way we will be able to write Midi to MemoryStream as an example and work on it directly in the memory instead of dumping the file to drive

MaKiPL:
I've made a pull request #499. I'm closing this issue

PR #499, opened 2019-06-25
Added MidiFile.Export() function overload supporting streams. Issue #498

MaKiPL:
Related to issue #498 
I added the support to MidiFile.Export() to work on Streams for example to supply MemoryStream instead of File I/O


MaKiPL:
I'm so sorry for such errors- this is an effect of corporate firewall and working from Notepad++

markheath:
thanks for this contribution. It has been re-implemented against main

PR #522, opened 2019-08-21
Add SimpleGateEffect ISampleProvider implementation

jizc:
I followed the `SimpleCompressorEffect` class as an example, and added a similar implementation for `SimpleGate`.

It would be nice to have this in the NAudio library since the `SimpleGate` class is `internal`.


markheath:
Thanks for this contribution and apologies for the delay. The plan is that NAudio 3 is going to have a broader selection of effects (currently work in progress on the naudio3-effects branch) and that will include a new `GateEffect` and `SimpleGate` has been removed

PR #534, opened 2019-09-23
Add MP3 frame duration property

AlFasGD:
https://stackoverflow.com/questions/383164/how-to-retrieve-duration-of-mp3-in-net/13269914#13269914

This very helpful answer has been up since '12, however such a workaround should be handled by the library. So I simply modified the code a little bit and put it in a property in MP3 frame. No idea if this is applicable in any other frame types, since I'm not too familiar with audio processing, however it should do the thing for the MP3 use case.


AlFasGD:
I've not really tested it at all and the code was copied from the attached SO answer which was accepted. The code was adjusted accordingly so that the code looks cleaner.

PR #545, opened 2019-10-13
add overload Transform method to operate on an array of samples

k2fc:


markheath:
Closing this as NAudio 3 has a Span-based, inplace overload for BiQuad filter. Thanks for your contribution

PR #547, opened 2019-10-16
Avoid to have an open handle of the input file

notjulian:


markheath:
thanks for PR, have made a similar change

PR #553, opened 2019-10-29
CPU intrinsics

macaba:
This pull request is here to generate some discussion about integrating the CPU intrinsics available in .NET Core 3.0 into NAudio by providing a code example of how it might be integrated into a class.

I benchmarked this pull request, and as expected, VolumeSampleProvider runs significantly faster:

image

Is this compatible with your vision for NAudio? 

Would you be interested in a larger pull request that uses intrinsics in a wide variety of places?

### DIFF
--- a/NAudio/Wave/SampleProviders/VolumeSampleProvider.cs
+++ b/NAudio/Wave/SampleProviders/VolumeSampleProvider.cs
@@ -1,4 +1,9 @@
-﻿namespace NAudio.Wave.SampleProviders
+﻿#if NETCOREAPP3_0
+using System.Runtime.Intrinsics;
+using System.Runtime.Intrinsics.X86;
+#endif
+
+namespace NAudio.Wave.SampleProviders
 {
     /// <summary>
     /// Very simple sample provider supporting adjustable gain
@@ -34,10 +39,14 @@
             int samplesRead = source.Read(buffer, offset, sampleCount);
             if (Volume != 1f)
             {
+#if NETCOREAPP3_0
+                FastPath(buffer, offset, sampleCount);
+#else
                 for (int n = 0; n < sampleCount; n++)
                 {
                     buffer[offset + n] *= Volume;
                 }
+#endif
             }
             return samplesRead;
         }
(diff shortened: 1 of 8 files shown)


NAudio/Wave/SampleProviders/VolumeSampleProvider.cs
```diff
@@ -46,5 +55,53 @@ public int Read(float[] buffer, int offset, int sampleCount)
...
+                    while (pCurrent < pEnd)
+                    {
+                        var input = Avx.LoadVector256(pCurrent);
+                        var output = Avx.Multiply(input, volume);
+                        Avx.Store(pCurrent, output);
+                        pCurrent += 8;
+                    }
+                }
+
+                if (Sse.IsSupported)
```
KvanTTT: Maybe `else if (Sse.IsSupported)`?

macaba: Current logic works optimally I'd say and is clean to read & easy to extend later.

Using edge case of 1005 samples:
On modern processors: 125 AVX, 1 SSE, 1 scalar. 
(With your suggestion, 125 AVX, 0 SSE, 5 scalar)
On old processors: 251 SSE, 1 scalar.
On arm processors: 1005 scalar.

macaba:
I've set up a proper benchmarking project in my repo to allow for impartial comparison. 

Benchmark for VolumeSampleProvider:

image

I'm initially focusing on the classes that I use personally so my next class to optimise is the BiQuadFilter.

markheath:
Haven't had a chance to look at this in detail yet (unfortunately, other projects taking priority at the moment), but just wanted to jump in and say I like the idea of providing optimised versions of some of these classes, and looks like you've achieved a really impressive  speedup there. 

One thing that I have been experimenting with is whether an NAudio 2.0 should fully embrace Span<T>. I made a quick proof of concept a while ago in [this repo](https://github.com/markheath/spantest).  With this approach ISampleProvider would look like [this](https://github.com/markheath/spantest/blob/master/ISampleProvider.cs)

If I do get time to create that sort of NAudio 2.0, then making use of new features like CPU intrinsics would definitely be of use. There are also some benefits of making use of the new compiler intrinsics to access the Calli instruction which would help resolve some threading headaches with the COM interop

For the existing NAudio 1.9, I'm happy to take PRs for any performance changes that don't make breaking changes to the public interface, and have been adequately tested.

macaba:
I would say the use of Span in the way you suggest for NAudio 2.0 is ideal. The existing offset and sampleCount simply become:

```
volumeSampleProvider.Read(buffer.Slice(offset, sampleCount));
```

Zintom:
This is still a very interesting prospect, I'd like to see these tests done in .NET 5 to see if there's even more of a performance bump.

Are we going to see something like this in the next version of NAudio?

macaba:
.NET 5 doesn't provide any additional gain in the case where CPU intrinsics are used.
(understandably - CPU instrinsics is the absolute fastest way to do this)

image

In the case of no CPU instrinsics being used, the act of simply moving to .NET 5 from a .NET Core runtime does not yield any improvement.

image

Zintom:
Still, CPU intrinsics appear to be a clear win even if there's no difference between 3.1/5.0

macaba:
Yep, another example of CPU intrinsic speed up is in #794 if you’re curious.

Issue #479, opened 2019-05-15
Is it possible to open a WAV file for append?


jfrank14:
As I wrote about in [this StackOverflow post](https://stackoverflow.com/questions/56149530/how-can-i-append-to-a-very-large-wav-file-with-naudio), my application needs to repeatedly open massive WAV files--as in >1GB--and this is causing performance problems, because the user has to wait 5 or 10 seconds after hitting record before she can start recording, the time needed to open a new stream and the existing file and copy the bytes from old to new.

I can't open the file stream in append mode, because this causes an error when Disposing the stream and seeking to the beginning to write the header.

Is there any way to *append* to such a WAV file, or some OTHER way to solve this problem?

markheath:
I'm afraid that NAudio itself doesn't have this capability, but for another project I made my own custom version of WaveFileWriter that let you open an existing file for appending to. It's not too much effort to make the modification

jfrank14:
Is that other project open source, or would you be able to suggest how one would do this modification? Thanks!

markheath:
Sorry, it's not open source. But I just passed in an existing file stream to the constructor, read the WAV header to get the WaveFormat, and then position to the end of the file.

PR #570, opened 2019-12-12
add append wav

mootensai:
this PR makes WaveFileWriter could append to already written wav


markheath:
Thanks for this contribution, but due to changes in main (e.g. support for RF64) this apporach to supporting append wouldn't actually work. We'd need to implement differently. There is an open issue still for appending to WAV #479 so any further discussion of this feature request can continue there

Issue #538, opened 2019-10-07
WaveOut and WaveIn support in netstandard?


jskeet:
I'm working on a [WPF application for my electronic drum set](https://jskeet.github.io/DemoCode/Drums/) and currently target .NET Core 3.0 and .NET 4.7.2.

While much of NAudio is available in the netstandard2.0 target, WaveIn and WaveOut aren't available - and even WaveInEvent/WaveOutEvent don't allow me to detect the number of devices or get their capabilities.

Is this a fundamental limitation within .NET Standard, or could it be addressed in a future release? Would adding a netstandard2.1 target help? Let me know if there's anything you'd like me to investigate.

sipsorcery:
+1 for `WaveOut`. 

For input devices `WaveInEvent.DeviceCount` is available in `netstandard2.0`. Unfortunately no matching method on `WaveOutEvent`.

jwosty:
This also applies to a lot of the Asio stuff (like AsioOut). I wonder if there's any reason this can't target .net standard (for windows), or even .net core? .net core 3 can actually build WPF apps on windows, and I believe that's what MS wants you to do from now on, considering that .Net Framework 4.8 is going to be the last of the .Net Frameworks.

See how AsioOut.cs is just remove from compilation for .Net standard (are there apis missing from .net standard or .net core that's causing it to fail compilation?): https://github.com/naudio/NAudio/blob/ec5266ca90e33809b2c0ceccd5fdbbf54e819568/NAudio/NAudio.csproj#L28

jwosty:
After some quick investigation, AsioDriver's usage of Microsoft.Win32.Registry seems to be the only thing stopping preventing successful compilation: https://github.com/naudio/NAudio/blob/89dd9528a79d559c80f2bfd4a91600ec9fde2050/NAudio/Wave/Asio/ASIODriver.cs#L33

I've found that Microsoft has released a .net standard (and friends) compatible package for this: https://www.nuget.org/packages/Microsoft.Win32.Registry. I'm really curious now. I'm going to see if I can get this part to work under .net standard

jwosty:
Now that #574 is merged, here is the situation:

- .Net Standard 2.0 supports ASIO
- .Net Core 3.0 supports WaveIn/WaveOut

I think this can be closed, because this should be sufficient for most use cases as new apps should probably be using .Net Core 3.0 anyway.

jskeet:
[...]

PR #574, opened 2019-12-30
Support ASIO and WAVEIN/WAVEOUT in .Net Core 3.0

jwosty:
Address #538 in spirit. Specifically, this PR brings the missing ASIO code into the .Net Standard build of NAudio, as well as the WaveIn and WaveOut code into the .Net Core 3.0 build.

The latter works because Windows Forms (used by WaveIn/WaveOut for message pumping: see [NAudio/Wave/MmeInterop/WaveWindow.cs](https://github.com/naudio/NAudio/blob/f5e7c103bb05d819ed16d379737bb804652ce2fd/NAudio/Wave/MmeInterop/WaveWindow.cs)) is now supported by .Net Core 3.0 but not .Net Standard.

This PR also modifies the WPF and WinForms sample apps to multi-target for .Net Framework and .Net Core 3.0 to demonstrate the targeting changes.


jwosty:
Build failure appears to suggest that CI needs to upgrade to .Net Core 3.0.

Molive-0:
Can confirm this build works and that the CI is at fault.

markheath:
thanks for this, and apologies for the delay - haven't had much time to work on NAudio recently. I just need to test and work out if I can get the Azure Devops build going. I think it might just be a case of needing to update the `azure-pipelines.yml` file to set `vmImage` to `windows-2019` as [documented here](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops#use-a-microsoft-hosted-agent)

Issue #581, opened 2020-01-21
UWP IWavePlayer cross-platform problems


neilt6:
I'm trying to use NAudio in a cross-platform Xamarin.Forms project, but I'm running into runtime errors in my UWP project when using `WasapiOutRT`. I've traced this back to UWP getting a slightly different definition for `IWavePlayer`:

```
#if WINDOWS_UWP
        /// <summary>
        /// Obsolete init method
        /// </summary>
        /// <param name="waveProvider"></param>
        /// <returns></returns>
        [Obsolete]
        System.Threading.Tasks.Task Init(IWaveProvider waveProvider);

        /// <summary>
        /// Initialise playback
        /// </summary>
        /// <param name="waveProviderFunc">Function to create the waveprovider to be played
        /// Called on the playback thread</param>
        void Init(Func<IWaveProvider> waveProviderFunc);
#else
        /// <summary>
        /// Initialise playback
        /// </summary>
        /// <param name="waveProvider">The waveprovider to be played</param>
        void Init(IWaveProvider waveProvider);

        /// <summary>
        /// The volume 
        /// 1.0f is full scale
        /// Note that not all implementations necessarily support volume changes
        /// </summary>
        float Volume { get; set; }
#endif
```

IMHO, an interface shouldn't have multiple definitions like this. I'd like to propose we standardize it to avoid problems in cross-platform scenarios.

neilt6:
@markheath I've managed to fix part of this problem with e069bf5. I'm interested in hearing feedback on how we could address the different `Init()` definitions.

neilt6:
See issue #585 for context on how I'm doing cross-platform with NAudio.

markheath:
Closing this old issue as the UWP support has been dropped in NAudio 3 (currently in preview) in favour of just targeting Windows. For cross platform you can look at the new NAudio.Alsa and NAudio.SoundFile libraries and see if they meet your needs.

PR #582, opened 2020-01-21
Added IWavePlayer.Volume to UWP

neilt6:
Added the Volume property to the UWP IWavePlayer interface to improve cross-platform compatibility. Issue #581.

### DIFF
--- a/NAudio/Wave/WaveOutputs/WasapiOutRT.cs
+++ b/NAudio/Wave/WaveOutputs/WasapiOutRT.cs
@@ -451,9 +451,9 @@ namespace NAudio.Wave
         }
 
         /// <summary>
-        /// Volume (not implemented)
+        /// Volume (not supported)
         /// </summary>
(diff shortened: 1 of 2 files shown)


NAudio/Wave/WaveOutputs/WasapiOutRT.cs
```diff
@@ -450,6 +450,11 @@ public PlaybackState PlaybackState
             }
         }
 
+        /// <summary>
+        /// Volume (not implemented)
+        /// </summary>
+        float IWavePlayer.Volume { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
```
markheath: Get should return 1.0f, Set should throw a `NotSupportedException`

neilt6: Agreed, fixed.

PR #587, opened 2020-02-03
fix CurrentTime not positioning to a multiple of BlockAlign

bentorkington:
setting CurrentTime of a WaveStream could result in the Position being set to something that is not a multiple of BlockAlign

### DIFF
--- a/NAudio/Wave/WaveStreams/WaveStream.cs
+++ b/NAudio/Wave/WaveStreams/WaveStream.cs
@@ -105,7 +105,10 @@ namespace NAudio.Wave
             }
             set
             {
-                Position = (long) (value.TotalSeconds * WaveFormat.AverageBytesPerSecond);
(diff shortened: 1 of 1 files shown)


markheath:
what situation were you getting this issue in? This is normally handled one layer up e.g. (WaveFileReader, Mp3FileReader)

bentorkington:
Hi Mark.

I'd only hit it in one of my own WaveStream subclasses, after refactoring it's callers from using Position (which I'd block aligned in the caller) to CurrentTime (which starting spitting garbage because my subclass relied on the caller doing the alignment).

It looks like I have misinterpreted [these docs](https://markheath.net/post/naudio-wavestream-in-depth), particularly the word _underlying_:

> If this is the case, be careful to ensure that the **underlying** stream is never set to a non block-aligned position. If it is, the resulting audio will be garbage.

I'd mistakenly interpreted this as a constraint of the WaveStream.Position property, and thought that the CurrentTime setter should respect this. Sorry, and thanks for such a great library.

PR #590, opened 2020-02-07
Add .NET Core 3.1 support

sirjeppe:
Seems 3.1 works just as well as 3.0.

### DIFF
--- a/NAudio/NAudio.csproj
+++ b/NAudio/NAudio.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="MSBuild.Sdk.Extras">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp3.0;net35;netstandard2.0;uap10.0</TargetFrameworks>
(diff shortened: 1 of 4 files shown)


jwosty:
I don’t believe this is necessary. In general, libraries should target the lowest runtime version that has all the features they want, which for the WaveOut code is .Net Core 3.1. Otherwise you’ll end up targeting .net core 3.0, 3.1, 3.1, 3.3.... etc when a target of 3.0 alone will allow it to be used on all of them. 

Microsoft has some useful guidelines: https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/cross-platform-targeting

markheath:
yes,  thanks for the contribution but there is no real benefit in adding 3.1 as an additional target. You can use NAudio just fine in 3.1 as is.

sirjeppe:
Oh, this is bad. I'm sorry for this! Visual Studio did not allow me to install the 1.10.0 version into a .NET Core 3.1 project for some reason, so I went ahead and made it 3.1 compatible and then I could install it. But now that I tested again to install 1.10.0 into my 3.1 project it worked flawlessly. Don't know what happened there...

Closing this!

Thank you for the quick (and kindly worded) feedback!

Issue #591, opened 2020-02-08
Split NAudio into multiple packages


jwosty:
After reading #184 I think it is time to take the .Net Standard spirit to NAudio, and split it into multiple packages that each target different platforms. The first step would be to identify the completely platform-independent interfaces and classes (for example IWavePlayer and WdlResamplingSampleProvider) and refactor them out into a new package, perhaps NAudio.Core, keeping the same namespaces for backward-compatibility. The next step would be to move the various platform/api-specific classes and interfaces into their own separate packages each; for example putting all the Asio stuff into a new package called NAudio.Asio.

This is a huge step forward for opening up the possibility for new platform implementations. For example, a motivated person could create macOS implementations of the NAudio API, and call it NAudio.AppleCoreAudio in the future. Here's a quick-n-dirty outline of the possible new package structure:

* NAudio.Core (.Net Standard 2.0)
  * Shared interfaces and functionality like IWavePlayer and WdlResamplingSampleProvider
* NAudio.Asio (.Net Standard) (Windows-only)
* NAudio.WaveOut (.Net Framework 3, .Net Core 3.0) (Windows-only)
* NAudio.WindowsCoreAudio (.Net Standard or whatever) (Windows-only)
* _NAudio.AppleCoreAudio (Xamarin.Mac)_
  * in the future, someone could certainly create this as a third-party library, or even submit a PR to the official one to implement this
  * macOS codecs and audio playback/recording APIs, probably based on AudioQueues

_(I could also imagine breaking things down further, like splitting playback and recording into different packages, though this may be overkill)_

Finally, we would also have to make sure to create a new NAudio package that includes no code but simply includes all the other packages directly, again for backward compatibility, as not to break anything that directly reference the "NAudio" package. Probably also mark it as deprecated.

I'm willing to take this on, as it is mostly dumb refactoring; although if someone beats me too this I won't be too upset :)

markheath:
This is something I've given quite a lot of thought to, and if there were to be a NAudio version 2 I'd definitely break it up into smaller chunks.

My ideas are similar with an NAudio.Core package that contains all the interfaces, plus all pure cross-platform .NET standard code (e.g. things like WaveFileReader and pure C# codec or effect implementations). 

[...]

PR #593, opened 2020-02-09
Split NAudio into multiple NuGet packages

jwosty:
#591 - .Net Standard-ify NAudio


jwosty:
Note that it's easier to review these changes in the desktop GitHub app, as it shows that almost all the changes are simple renames

jwosty:
@markheath would you be willing to take a PR like this, or should I try to split it up into smaller chunks? Renaming NAudio/ to NAudio.Core/ is generating a huge changeset which is probably 95% of this PR.

jwosty:
@markheath I would love to know your thoughts

tedvanderveen:
This is great @jwosty!

jwosty:
@tedvanderveen If anything I may split this effort off into a fork, since I would like to use this functionality sooner rather than later

markheath:
I like the idea of breaking NAudio up into packages. Can you just quickly summarise what went where because GitHub is doing an awful job of showing the diffs in a consumable way? In particular did you move DMO and MediaFoundation in with WASAPI?

markheath:
I've been having an experiment and have pushed a [v2 branch](https://github.com/naudio/NAudio/tree/v2), as a starting point for how things might be divided. There is still work to do on cleaning up namespaces, but might be worth you having a look and seeing if that breakdown would work for you.

The other big thing I'd love to get into a v2 of NAudio would be to make it based on Span<T>

DualBrain:
With regard to "span", please keep in mind us VB developers that also consume this project.

markheath:
Is there a problem using Span from VB?

DualBrain:
Obviously if the span(s) is/are kept completely internal to the implementation... then there's no problem.  However, VB contains little to nothing to create, support or interact with spans.  For example...

Dim arraySpan = New Span(Of Byte)(array)

Creates a (BS!) error stating 'Span(Of Byte)' is obsolete: 'Types with embedded references are not supported in this version of your compiler.'

"obsolete" isn't really the right word in that error.

So if there is / are any requirements to initialize, interact with spans from "other languages"... the advice would seem to be that "spans" are something to only be used within the implementation but not exposed on/through the API layer.

jwosty:
Well, speaking of multiple languages, I'm using NAudio from F#, and F# can consume Spans just fine (though there's a little roughness around the edges when it comes to certain parts). Yikes. That's horrifying that VB doesn't support them very well. I would normally ask why people are using this library anymore from VB, but given how long NAudio has been around, I get it...

DualBrain:
"I would normally ask why people are using this library anymore from VB"

I, for one, think NAudio is AWESOME and recommend it to everyone I interact with when it comes to working with audio in .NET.

jwosty:
@markheath Looking at this PR again, it seems I pretty much just made the following splits (by splitting some things into new csprojects and therefore creating some new packages):

* NAudio.Core - as far as I remember, it contains the core interfaces and everything that could be shared across platforms (so that in the future someone [probably me :)] could come along and write a macos implementation for things)
* NAudio.Asio - all the Asio implementations (pretty much just AsioOut and everything ASIO-specific that it needs)
* NAudio.Wasapi - all the Wasapi implementations (again, pretty much just WasapiOut and its non-shared dependencies)

The end result is that in this PR now we have these new NAudio.Core, NAudio.Asio, and NAudio.Wasapi projects separated into their own folders. Which is why Github is generating the awful diff as you can see. I didn't change any namespaces or any code at all, just drew some extra lines for new packages. Pretty much just moved files around, messed with the build a little bit, and made some new csprojects.

@DualBrain yes, I love this library too and I want to see it flourish just like you! @markheath has done a phenomenal job with this beast of a library and his efforts are not going for granted. I also don't want to come off like I'm dissing VB, haha. VB programmers are people too ;)

DualBrain:
"I also don't want to come off like I'm dissing VB, haha."

Although I don't use F#... I am a HUGE fan of the F# community.  I'm doing what I can to help nudge the VB community to be inspired by the F# community to get more involved in our own future.

For example, I just kicked off a project to "take ownership" of the "Visual Basic Runtime" from a more community point of view; in one area particular, I'm planning on utilizing NAudio for the My.Computer.Audio functionality (when running on Windows) - yes, one of the goals is to take the runtime cross platform as I'm also interested in Linux and RasPi.

jwosty:
@markheath I'm looking at your v1 branch and yeah, it looks like you've done pretty much what I've done here, but you've done more packages as well. Seems like we're on the same page here!

@DualBrain yes! As time goes on it really seems like the best way to get stuff you want in your language. If the OSS community didn't push F# as hard as they have been by digging into the guts of stuff themselves, F# wouldn't be where it is today. The Microsoft F# team (who are all very passionate and dedicated individuals of course) only has so much manpower and MS doesn't grow them very much! Your projects sound very interesting, anyway!

I, personally, am using NAudio for a cross-platform desktop application that will be used in the live performance industry. That's why I'm so invested in the macos aspect of stuff.

jwosty:
Perhaps it may be more relevant to move this discussion to #591 as it seems this draft PR might be unnecessary at this point, since the work is being done in a different branch

markheath:
I'm going to close this PR, as I've done it in a slightly different way on the [v2 branch](https://github.com/naudio/NAudio/tree/v2)
Many thanks for your work on this which prompted me to finally get this done

PR #599, opened 2020-02-22
[ASIO] Added DriverResetRequest event 

tebjan:
This occurs when settings are changed by the user, e.g. in the ASIO control panel.
Many professional soundcards have a dedicated control panel, with this event it is possible to react to user changes.


markheath:
looks good, thanks

PR #600, opened 2020-02-22
[ASIO] Exposing IsSampleRateSupported from asio driver

tebjan:
Just allowing access to this helper method.


markheath:
looks good, thanks

PR #602, opened 2020-02-28
Added CachedSound constructor for playing resource wav files

YJ15:
Added CachedSound constructor for playing resource wav files
I played the wav file of Properties.Resource using the following source code:

```csharp
public CachedSound(Stream sound) 
{
        using (var audioFileReader = new WaveFileReader(sound))
        {
            WaveFormat = audioFileReader.WaveFormat;
            var sp = audioFileReader.ToSampleProvider();
            var wholeFile = new List<float>((int)(audioFileReader.Length / 4));
            var sourceSamples = (int)(audioFileReader.Length / (audioFileReader.WaveFormat.BitsPerSample / 8));
            var sampleData = new float[sourceSamples];
            int samplesread;
            while ((samplesread = sp.Read(sampleData, 0, sourceSamples)) > 0)
            {
                wholeFile.AddRange(sampleData.Take(samplesread));
            }
            AudioData = wholeFile.ToArray();                
        }
}
```

### DIFF
--- a/NAudioWpfDemo/FireAndForgetPlayback/CachedSound.cs
+++ b/NAudioWpfDemo/FireAndForgetPlayback/CachedSound.cs
@@ -1,6 +1,7 @@
 ﻿using System.Collections.Generic;
 using System.Linq;
 using NAudio.Wave;
+using System.IO;
 
 namespace NAudioWpfDemo.FireAndForgetPlayback
 {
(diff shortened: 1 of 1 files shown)


YJ15:
Ok. I will modify this soruce code.

YJ15:
"using System.IO;" Is missing.
I modify it

sampletext32:
I think it's not good write to a list, it's better to write to a buffer(array) with some offset.

jcummings2:
I needed this change as well, so I adapted it for the new layout in pull request #963.

markheath:
thanks for this contribution. It has now been implemented in NAudio 3

PR #661, opened 2020-07-03
Fix silent failure in WasapiCapture/WasapiOut

roadz99:
When using Wasapi with event sync, if the event was not set then
the audiClient was never queried. If the audioClient has failed for
some reason then the WasapiCapture/WasapiOut will now
notice this and raise PlaybackStopped/RecordingStopped event.


markheath:
thanks

PR #662, opened 2020-07-06
MixingWaveProvider32 - do not add same input stream many times

DavidMartynWood:


NAudioTests/WaveProviders/MixingWaveProvider32Tests.cs
```diff
@@ -0,0 +1,24 @@
+﻿using NAudio.Wave;
+using NUnit.Framework;
+
+namespace NAudioTests.WaveProviders
+{
+    [TestFixture]
+    public class MixingWaveProvider32Tests
+    {
+        [Test]
+        public void CannotAddSameInputInputProviderSecondTime()
```
sampletext32, line 10: I believe this is a typo

```suggestion
        public void CannotAddSameInputProviderSecondTime()

```

markheath:
Thanks for this contribution.

`MixingWaveProvider32` has now been deleted as it is obsoleted by `MixingSampleProvider`, but I have applied this guard to `MixingSampleProvider`

PR #677, opened 2020-08-06
Sign NAudio assembly

ermshiperete:
This change creates a strong-named version of NAudio.dll. While this
doesn't have any benefits in regards to security it helps those projects
that (for whatever reason) are required to create strong-named assemblies
because all dependencies have to be strong-named as well.


markheath:
The new NAudio 2 (currently in beta) will used strong named assemblies. Please let me know if that meets your needs

ermshiperete:
Yes, that helps. Thanks!

But it still would be good to have it available in v1 as well - I haven't checked what changed between v1 and v2, but the major version number change suggests that there are breaking changes that not every project can make immediately. And v2 is still in beta, whereas it should be possible to release a new stable v1 package right away (with the only change being that it is now signed).

markheath:
NAudio v2 will be highly compatible. All the same classes, with same names and same namespaces, just organized into different assembly groupings (which are strongly typed).

ermshiperete:
OK, sounds good. I'll close this one then.

PR #714, opened 2020-11-30
Added readonly property "BufferedWaveProvider" to WaveInProvider

Mostlypyjamas:
Currently there is no way to determine how many (if any) bytes have been added to the internal Buffer by the WaveInProvider.

This PR exposes the internal BufferedWaveProvider via a ReadOnly property so that consumers can determine how many bytes are available, as well as allowing control over the internal ReadFully property if required.

### DIFF
--- a/NAudio.Wasapi/MediaFoundationEncoder.cs
+++ b/NAudio.Wasapi/MediaFoundationEncoder.cs
@@ -57,13 +57,11 @@ namespace NAudio.Wave
                     throw;
                 }
             }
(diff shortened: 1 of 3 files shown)


Mostlypyjamas:
I'm not sure why it's failing the checks, but seems to be a Nuget issue in the azure pipeline?

Redyoung49:
> I'm not sure why it's failing the checks, but seems to be a Nuget issue in the azure pipeline?

same question

PR #716, opened 2020-12-08
cuelist support standard UTF8 character

Redyoung49:
this code ` Label = Regex.Replace(label, @"[^\u0000-\u00FF]", "");` would replace some utf8 character like chinese character[\u0080-\u07FF] , so I delete it and recalculate the length in `internal byte[] GetRiffChunks()`

### DIFF
--- a/NAudio/Wave/WaveStreams/CueList.cs
+++ b/NAudio/Wave/WaveStreams/CueList.cs
@@ -30,11 +30,7 @@ namespace NAudio.Wave
         public Cue(int position, string label)
         {
             Position = position;
-            if (label == null)
-            {
-                label = "";
-            }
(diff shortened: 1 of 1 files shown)


markheath:
WAV is a very old file format. Are you sure that if you write UTF encoded data into a cue list, other applications will be able to read it correctly? What have you tested with?

Redyoung49:
> WAV is a very old file format. Are you sure that if you write UTF encoded data into a cue list, other applications will be able to read it correctly? What have you tested with?

I test it in adobe audition,and ues audition write marker and read it.this cuelist code use in my project. my project need read/ write/modify cue in wav file.
but now , I have to rewrite a new CueWaveFileReader/CueWaveFileWriter to solve this problem

PR #730, opened 2021-01-18
Incoming MIDI sysex support

brianavid:


markheath:
don't worry too much about the checks - the CI is broken on master - and fixed in the V2 branch (hopefully merging soon)

NAudio/Midi/MidiIn.cs
```diff
@@ -48,6 +57,16 @@ public MidiIn(int deviceNo)
         /// </summary>
         public void Close() 
         {
+            //  Free up all created and allocated buffers for incoming Sysex messages
```
markheath, line 60: this code should move into the `Dispose` method so that it gets cleared up if you use it with a `using` block

brianavid: I agree - I had just re-read your Midi doc, which says to use Dispose(), rather than Close() and thought that. I never even considered the issue of "using". Would you just do that or to I need to update the PR in some way? Also, have I made the change on the wrong branch? I notice that my ChannelAftertouch fix went on the V2 branch. The use of branches isn't very clear I'm afraid.

markheath: You can push new changes to your branch and this PR will automatically update itself.
NAudio has had just one branch for years, but the v2 stuff is being done in a separate branch and is going to be imminently released (when I have time to fix). I've been merging changes from v1 to v2 in the meantime.

brianavid: OK - new commit pushed, moving the cleanup code to Dispose(). I see I have a lot to learn about Git before I attempt any more contributions to public open source projects!!

brianavid:
This PR is not yet ready. I have now realised that it is preventing a clean shutdown with MidiInterop.midiInClose(hMidiIn) returning 65 which I think is MIDIERR_STILLPLAYING. Investigating ...

markheath:
OK cool, keep me posted

brianavid:
Got it. midiInClose won't work if there are Sysex buffers registered. A call to midiInReset() is needed to flush these out and the callback should not re-add the buffers if there is no registered event handler. I've also moved the whole closure code to the virtual Dispose() for widest coverage.

brianavid:
I've been running my build with this change now for 5 days and have had no issues. I'm happy if and when you want to merge.

brianavid:
I see that there has been significant activity on V2 in the last couple of weeks and I PR'd this change on the master branch. Should I fetch and re-do it again on V2?? Would that make things easier?

markheath:
sorry for the delay getting back to you. I've not had much time for NAudio recently. Yes, it would probably be easiest to resubmit these changes against V2. Lots of things have moved to new places, but on the whole the code itself hasn't changed much so shouldn't be too hard to achieve

brianavid:
Ported the changes to V2 - trivial - nothing needed to be altered. This port now has its own PR. So this old PR can be abandoned.

markheath:
closing in favour of #748

PR #733, opened 2021-01-26
Fix CircularBuffer Backward Bug

euojeff:
We run into this bug when we need to create an Audio Cache using CircularBuffer.
I need to be able to return the Buffer, already read, to be able to take advantage of information that has already been written. This is great for creating a cache of realtime processed audio.

I fix this litle bug and would like to contribute. Unit tests have been added for this class.

Before:
backError

After:
backTestOk

ps: We do not check if the requested backward is greater than the buffer in order to not overcharge. Who uses this class that should do this treatment.


euojeff:
I had put the test in the wrong class. I moved to the correct class.

tests

markheath:
Not sure I'd call this a bug - the `CircularBuffer` is not designed to go backwards. I'd recommend using your own custom buffer if you need that behaviour.

PR #741, opened 2021-02-11
Play audio with MediaFoundation

ISDHN:
Add function:
    Play audio with MediaFoundation


markheath:
Thanks, looks like a nice addition to the library although I will need some time to read the code and maybe try it.
Can you tell me a bit about what platforms you have tested this on.
Also, ideally can you use four spaces for indentation instead of tabs, which fits with the code style elsewhere in NAudio and will prevent some false diffs in the code review.

ISDHN:
It's a honor for me to contribute to the NAudio project,and I'm so surprise at the acceptance of my pull request.
As for the question,I've test this idea on Window 7 and Window 10,and it runs well.
Besides,thanks to your advice.I will change the indentation from tabs to four spaces





------------------&nbsp;原始邮件&nbsp;------------------
发件人:                                                                                                                        "naudio/NAudio"                                                                                    <notifications@github.com&gt;;
发送时间:&nbsp;2021年2月11日(星期四) 晚上11:08
收件人:&nbsp;"naudio/NAudio"<NAudio@noreply.github.com&gt;;
抄送:&nbsp;"星辰五岳"<864593988@qq.com&gt;;"Author"<author@noreply.github.com&gt;;
主题:&nbsp;Re: [naudio/NAudio] Play audio with MediaFoundation (#741)





 
Thanks, looks like a nice addition to the library although I will need some time to read the code and maybe try it.
 Can you tell me a bit about what platforms you have tested this on.
 Also, ideally can you use four spaces for indentation instead of tabs, which fits with the code style elsewhere in NAudio and will prevent some false diffs in the code review.

markheath:
Have been revisiting old PRs and looking at this one. My main question at the moment is what value does this bring given that NAudio has WASAPI which is the main recommended way to play audio on Windows (and ASIO for low latency)?

Are there capabilities that MF playback offers that WASAPI doesn't? Or situations in which is would be the better choice?

markheath:
And in what way is `MediaFoundationProvider` from this PR different from `MediaFoundationReader` which can also read from a URL?

mdcdi1315:
> Are there capabilities that MF playback offers that WASAPI doesn't? Or situations in which is would be the better choice?

Per my findings, the only situation that MF playback is better than WASAPI is adding multiple `IWaveProvider` instances per player instance. Other than that, WASAPI offers all the other things that MF playback does offer, and it is sufficient for typical usage scenarios. Even if there is the need for multiple wave providers, `ConcatenatingSampleProvider` is enough.

markheath:
Agreed, I think we will leave this out of NAudio unless the original author knows a reason why this is particularly necessary.
With NAudio the goal is not to ship wrappers for every possible audio API, but to cover the most important ones. For Windows that's WASAPI for record/playback and MF for format handling. And ASIO for low-latency record/playback scenarios.

PR #754, opened 2021-03-19
Fix WaveStream.CurrentTime

tweigt:
`WaveStream.CurrentTime` can set `WaveStream.Position` to a fraction of `WaveStream.BlockAlign`.
This fix makes the setter use `SampleRate` and `BlockAlign` to make sure Position is set to a value that corresponds to a Sample boundary.

This should not be a breaking change, as the current behavior is clearly buggy.

### DIFF
--- a/NAudio.Core/Wave/WaveStreams/WaveStream.cs
+++ b/NAudio.Core/Wave/WaveStreams/WaveStream.cs
@@ -105,7 +105,7 @@ namespace NAudio.Wave
             }
             set
             {
-                Position = (long) (value.TotalSeconds * WaveFormat.AverageBytesPerSecond);
+                Position = (long) (value.TotalSeconds * WaveFormat.SampleRate) * BlockAlign;
             }
         }
 


Zintom:
This is a breaking change, although yes it does fix "buggy behaviour", some apps may rely on this behaviour to work.

tweigt:
Ok, then at least for now there should be a hint in the summary comment, that setting a time with fractional seconds can cause the audio to be corrupted, if `TotalSeconds * SampleRate` is not an integer.

When I first tried out the library, it took me way longer than I would have liked to figure out why the audio was playing fine sometimes and more often it was basically white noise, when I skipped through it with a slider.

Should I make a separate pull request for that?

Zintom:
> When I first tried out the library, it took me way longer than I would have liked to figure out why the audio was playing fine sometimes and more often it was basically white noise, when I skipped through it with a slider.

That's very understandable, the XML documentation is quite lacking I'd say.

> Should I make a separate pull request for that?

Yeah that makes sense.

markheath:
Thanks for this contribution. The underlying issue is now fixed in main

PR #756, opened 2021-03-23
Release asio driver if init was with error

piksaevalex:
I use NAudio to get sound from Dante devices via AsioDriver. 
I found a problem with the driver: If i init AsioDriverExt when device unavailable, then future connect attempts return exception code like Dante Visual Soundcard is already in use. My code resolve problem.


piksaevalex:
Hi @markheath, could you please look at my PR?

Issue #758, opened 2021-03-24
AudioSessionControl property


neumatho:
Hi,

I need the AudioSessionControl property to be implemented in AudioClient.cs for WASAPI. I need it to be able to detect default endpoint switch as shown in this example: [WASAPIRenderer.cpp](https://github.com/microsoft/Windows-classic-samples/blob/master/Samples/Win7Samples/multimedia/audio/RenderSharedEventDriven/WASAPIRenderer.cpp). Look at function InitializeStreamSwitch, first line.

I have already implemented it by taking your code and embedded it, but I would like to use the NuGet packet instead :-)

Here is the code I have written:

```
        private AudioSessionControl audioSessionControl;

        /// <summary>
        /// Gets the AudioSessionControl service
        /// </summary>
        public AudioSessionControl AudioSessionControl
        {
            get
            {
                if (audioSessionControl == null)
                {
                    var audioSessionControlGuid = new Guid("F4B1A599-7266-4319-A8CA-E70ACB11E8CD");
                    Marshal.ThrowExceptionForHR(audioClientInterface.GetService(audioSessionControlGuid, out var audioSessionControlInterface));
                    audioSessionControl = new AudioSessionControl((IAudioSessionControl)audioSessionControlInterface);
                }
                return audioSessionControl;
            }
        }
```

and in Dispose():

```
                if (audioSessionControl != null)
                {
                    audioSessionControl.Dispose();
                    audioSessionControl = null;
                }
```

-Thomas

markheath:
NAudio already supports what you need here, just not as a property on `AudioClient` itself.

If you want the session-disconnect signal (the `OnSessionDisconnected` path that the stream-switch sample uses), use `AudioSessionControl`, which NAudio exposes via the session manager:

```csharp
var device = new MMDeviceEnumerator().GetDefaultAudioEndpoint(DataFlow.Render, Role.Multimedia);
var session = device.AudioSessionManager.AudioSessionControl;
session.RegisterEventClient(myHandler); // myHandler : IAudioSessionEventsHandler
// myHandler.OnSessionDisconnected(AudioSessionDisconnectReason reason) fires on a switch/removal
```

If what you actually care about is the *default* render endpoint changing, the more direct approach is to register an `IMMNotificationClient` and handle `OnDefaultDeviceChanged`:

```csharp
var enumerator = new MMDeviceEnumerator();
enumerator.RegisterEndpointNotificationCallback(myNotificationClient);
```

[...]

PR #766, opened 2021-04-20
Added AudioSessionControl property.

neumatho:
Created this pull request for issue #758. Here is a resume:

"I need the AudioSessionControl property to be implemented in AudioClient.cs for WASAPI. I need it to be able to detect default endpoint switch as shown in this example: WASAPIRenderer.cpp. Look at function InitializeStreamSwitch, first line."

-Thomas

### DIFF
--- a/NAudio.Wasapi/CoreAudioApi/AudioClient.cs
+++ b/NAudio.Wasapi/CoreAudioApi/AudioClient.cs
@@ -17,6 +17,7 @@ namespace NAudio.CoreAudioApi
         private AudioClockClient audioClockClient;
         private AudioStreamVolume audioStreamVolume;
         private AudioClientShareMode shareMode;
+		private AudioSessionControl audioSessionControl;
 
         public AudioClient(IAudioClient audioClientInterface)
         {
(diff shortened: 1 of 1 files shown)


JacqiPan:
The AudioSessionControl is already implemented within  [AudioSessionControl.cs](https://github.com/naudio/NAudio/blob/master/NAudio.Wasapi/CoreAudioApi/AudioSessionControl.cs) and can be accessed within [AudioSessionManager.cs](https://github.com/naudio/NAudio/blob/master/NAudio.Wasapi/CoreAudioApi/AudioSessionManager.cs), or _MMDevice.AudioEndpointVolume.AudioSessionControl_.

The [WASAPIRenderer.cpp](https://github.com/microsoft/Windows-classic-samples/blob/master/Samples/Win7Samples/multimedia/audio/RenderSharedEventDriven/WASAPIRenderer.cpp) uses [RegisterEndpointNotificationCallback ](https://docs.microsoft.com/en-us/windows/win32/api/audiopolicy/nf-audiopolicy-iaudiosessioncontrol-registeraudiosessionnotification)to receive and handle (e.g. OnSessionDisconnected) notifications from the endpoint device, which can be achieved with NAudio by using _MMDevice.AudioEndpointVolume.AudioSessionControl.RegisterEventClient()_

Hope this helps

neumatho:
Thanks, I may have overlooked the AudioSesssionManager. I now uses that to get the AudioSessionControl and it work as it should.

-Thomas

markheath:
Seems like we don't need this then. Closing this

PR #769, opened 2021-04-22
Added EndpointFormFactor enumeration

JacqiPan:
Added the **_EndpointFormFactor_** enumeration. The constants in this enumeration are the values that can be assigned to the _**PKEY_AudioEndpoint_FormFactor**_ property, as described in https://docs.microsoft.com/en-us/windows/win32/api/mmdeviceapi/ne-mmdeviceapi-endpointformfactor

This allows for the ability to determine whether your device is set to one of the following:
- RemoteNetworkDevice
- Speakers
- LineLevel
- Headphones
- Microphone
- Headset
- Handset
- UnknownDigitalPassthrough
- SPDIF
- HDMI
- UnknownFormFactor

```csharp
    // Example use case:
    
    public EndpointFormFactor FormFactor(MMDevice device)
    {
            if (device.Properties.Contains(PropertyKeys.PKEY_AudioEndpoint_FormFactor))
            {
                return (EndpointFormFactor)device.Properties[PropertyKeys.PKEY_AudioEndpoint_FormFactor].Value;
            }
            else
                return EndpointFormFactor.UnknownFormFactor;        
    }
```

### DIFF
--- a/NAudio.Wasapi/CoreAudioApi/AudioClient.cs
+++ b/NAudio.Wasapi/CoreAudioApi/AudioClient.cs
@@ -252,7 +252,7 @@ namespace NAudio.CoreAudioApi
             {
                 return false;
             }
-            if (hresult == (int)AudioClientErrors.UnsupportedFormat)
+            if (hresult == AudioClientErrorCode.UnsupportedFormat)
             {
                 // documentation is confusing as to what this flag means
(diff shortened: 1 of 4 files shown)


JacqiPan:
Will move the EndpointFormFactor enumeration to a new branch.

PR #779, opened 2021-05-23
Add Stream support to MediaFoundationEncoder

aianlinb:


markheath:
Sorry for the very long delay getting round to this, but thanks very much, this is a great addition to the library

PR #794, opened 2021-06-12
Added NAudio.Core Benchmark project

macaba:
This is the beginning of a benchmarking project for NAudio.Core. The intention is to provide a way for developers to easily show the results of any performance improvements.

As an example, using the performance improvement shown in pull request #676 

Before:
image

After:
image

A very worthwhile improvement across all the runtimes, though that was not obviously shown within the pull request.


macaba:
I added a benchmark for a CPU instrinsic version of VolumeSampleProvider that yields a 10x performance improvement

image

macaba:
I vectorized the memory copy within WaveToSampleProvider to provide additional improvement:

image

To summarize (.NET 5.0):
Original: 10.45us
Pull request #676: 4.22us
Vectorized: 0.36us

The next step is to create unit tests for the classes using CPU intrinsics to be ensure the data copy is correct and prevent regressions.

PR #802, opened 2021-06-28
Added a Peek function to Circular Buffer and BufferedWaveProvider

Zintom:
Recently I've required the ability to 'peek ahead' at data in a `BufferedWaveProvider`, i.e read data without advancing the read position. At present there are no functions to do this.

I have implemented the `Peek` function on both the `CircularBuffer` and the `BufferedWaveProvider` and have also created a test to ensure that everything works as it should. I have also added rich documentation to each of the new methods.

This PR does not modify the existing public API methods, it only adds to them, so does not pose any risk when merging.

I look forward to discussing this!

### DIFF
--- a/NAudio.Core/Utils/CircularBuffer.cs
+++ b/NAudio.Core/Utils/CircularBuffer.cs
@@ -97,6 +97,47 @@ namespace NAudio.Utils
         }
 
         /// <summary>
+        /// Peeks ahead at data in the buffer without advancing the read position.
+        /// </summary>
+        /// <param name="peekOffset">The offset from the current read position where peeking will begin.</param>
+        /// <param name="data">The buffer we will read into.</param>
+        /// <param name="offset">The offset into the <paramref name="data"/> buffer.</param>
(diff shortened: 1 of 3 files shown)


NAudio.Core/Utils/CircularBuffer.cs
```diff
@@ -96,6 +96,47 @@ public int Read(byte[] data, int offset, int count)
...
+        /// <returns>The number of bytes actually read.</returns>
+        public int Peek(uint peekOffset, byte[] data, int offset, int count)
+        {
+            lock (lockObject)
+            {
+                int tmpReadPosition = readPosition + (int)peekOffset;
+                tmpReadPosition %= buffer.Length;
+
+                // The same implementation as Read however we use a local temporary read position variable.
+                if (count > byteCount)
```
Zintom, line 115: The implementation of the `Peek` method is actually almost identical to the `Read` method above it. The only difference being that the `Peek` doesn't modify the `readPosition` field, it creates a local temporary variable to log the read position. Also we don't increment the `byteCount` before we return to the caller.

PR #818, opened 2021-08-03
Fixed #801 - WaveViewer flipped vertically

### LINKED ISSUE #801: Waveform drawn by WaveViewer.cs is flipped vertically
From what I can tell, the drawing of the WaveViewer control is flipped vertically.  The GDI coordinate system places the origin point at the top-left corner, rather than in the center like Cartesian.  From the audio data, negative samples should plot down from the Cartesian x-axis, and positive up.

However, the WaveViewer calculates a "lowPercent" and a "highPercent", and then calculates the Y coordinate based on the control's height.  But, since "lowPercent" is calculated from negative samples, this means its Y coordinate is smaller than the "highPercent" Y coordinate.  Because in GDI the origin is in the top-left corner, this means that "low" is plotted above "high", resulting in a vertical-flipped image.

A solution is to use a properly configured transform RectangleF applied to the GDI graphics.  From there, the samples can be plotted directly without needing to calculate percentages.

vbasiccd:
Modified the OnPaint method to use a transform
rectangle for coordinate mapping. Fixes #801.

### DIFF
--- a/NAudio.WinForms/Gui/WaveViewer.cs
+++ b/NAudio.WinForms/Gui/WaveViewer.cs
@@ -5,6 +5,7 @@ using System.Drawing;
 using System.Data;
 using System.Windows.Forms;
 using NAudio.Wave;
+using System.Drawing.Drawing2D;
 
 namespace NAudio.Gui
 {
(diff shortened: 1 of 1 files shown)


markheath:
Thanks for this contribution. Has been implemented against main in PR #1304

PR #823, opened 2021-08-13
Fixed issue where BinaryReader.PeekChar would fail

LarsenLP:
Fixed issue where BinaryReader.PeekChar would fail, since bytedata is not UTF8 content

### DIFF
--- a/NAudio.Core/FileFormats/Wav/WaveFileChunkReader.cs
+++ b/NAudio.Core/FileFormats/Wav/WaveFileChunkReader.cs
@@ -88,9 +88,9 @@ namespace NAudio.FileFormats.Wav
                     }
                     if (storeAllChunks)
                     {
-                        if (chunkLength > Int32.MaxValue)
-                            throw new InvalidDataException(string.Format("RiffChunk chunk length must be between 0 and {0}.", Int32.MaxValue));
-                        riffChunks.Add(GetRiffChunk(stream, chunkIdentifier, (int)chunkLength));
+                        if (chunkLength > UInt32.MaxValue)
+                            throw new InvalidDataException(string.Format("RiffChunk chunk length must be between 0 and {0}.", UInt32.MaxValue));
+                        riffChunks.Add(GetRiffChunk(stream, chunkIdentifier, chunkLength));
                     }
                     stream.Position += chunkLength;
                 }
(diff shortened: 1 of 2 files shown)


markheath:
PeekChar and ReadByte behave differently. One advances the stream and the other doesn't. And ReadByte will return -1 if it reaches the end of stream, so I don't think checking for 0 is correct.

Do you have an example failing WAV file that can't be handled?

LarsenLP:
Hi Mark,

Thanks for your quick reply!

I totally agree that PeekChar and ReadByte behaves differently.

The issue with PeekChar is that the BinaryReader by default is created with encoding set to UTF8. So when using PeekChar, it will try to peek the next character from the reader, but this of course requires that the byte read from the stream can be successfully parsed as an UTF8 character. And when dealing with raw binary data, this is unfortunately not always the case.

The most optimal solution would be if the dotnet team had included a PeekByte function. That would have solved everything.

So my suggested way around this is to actually read the next byte from the stream using ReadByte:

- If the byte read is a pad byte (0) everything is good. The pad byte is consumed by ReadByte and the stream position is correctly  advanced with +1, so the next chunk read does not start with the pad byte.
- If it is not a pad byte, the position of stream will be set back one byte by "stream.Position--;", so that byte will be included in the next chunk read.

What are your thoughts on this?

I have created a sample app including audio files to reproduce the issue: [https://www.dropbox.com/s/ot693uy3hvclri9/NAudioPeekCharReproduction.zip?dl=0](https://www.dropbox.com/s/ot693uy3hvclri9/NAudioPeekCharReproduction.zip?dl=0)

One issue coming to mind: It might be worth checking for end of the stream before calling ReadByte...

Best,
Lars Peter

LarsenLP:
Closing this PR since I messed up my fork by committing another fix to the master branch. The two fixes are now on separate branches and submitted as separate pull requests.

New PR for "Fixed issue where BinaryReader.PeekChar would fail": [https://github.com/naudio/NAudio/pull/828](https://github.com/naudio/NAudio/pull/828)

New PR for "Fixed datatype of chunks from Int32 to UInt32": [https://github.com/naudio/NAudio/pull/827](https://github.com/naudio/NAudio/pull/827)

PR #881, opened 2022-02-28
Update AiffFileWriter.cs

JoeGuetz:
dataChunkSize needs to be defined bigger than int, else the multiplication in function UpdateCommChunk will return negative value if the product of the multiplication exceeds Int.MaxValue


markheath:
sorry for the delay, and thanks for your contribution

Issue #908, opened 2022-06-04
bug: infinite callback loop when disposing an instance having a sysex event handler registered.


flyingoverclouds:
The bug occurs when you create sysex buffers AND have defined a SysexMessageReceived handler.

```csharp
var midiOut = new MidiOut(devMidiOut);
var midiIn = new MidiIn(devMidiIn);
midiIn.MessageReceived += (object? sender, MidiInMessageEventArgs e) =>
{
    Console.WriteLine($"IN : Channel={e.MidiEvent.Channel}  Command={Enum.GetName(e.MidiEvent.CommandCode)}");
};
midiIn.ErrorReceived += (object? sender, MidiInMessageEventArgs e) =>
{
    Console.WriteLine($"ERROR IN : Channel={e.MidiEvent.Channel}  Command={Enum.GetName(e.MidiEvent.CommandCode)}");
};

midiIn.CreateSysexBuffers(20, 2);
midiIn.SysexMessageReceived += (object? sender, MidiInSysexMessageEventArgs e) =>
{
    Console.WriteLine($"SYSEX IN : size={e.SysexBytes.Length}    ts={e.Timestamp}");
};

midiIn.Start();

Console.WriteLine("Use the selected midi IN device to see Midi Message informations. PRESS ENTER TO TERMINATE");
Console.ReadLine();

midiIn.Stop();
midiIn.Dispose(); // <== INFINITE LOOP OF SysexMessageReceived
midiOut.Dispose();
```

When the Dispose(bool) Method call the midiInReset method, a callback with a LONGDATA message having size to zero is made for each allocated sysexbuffer.  
In the In that case, the buffer should not be re-added (midiInAddBuffer). if readded, the callback occurs again and start an inifinite loop of callback/addBuffer.
In the same way, The c# event should not be triggered for a zero length sysex message. 

A workaround on the user code is to remove each event handler from the SysexMessageReceived event before disposing the instance. But this workaround is not possible if you are using the lamba syntax. And the bug is still present if you don't explicitly call Dispose before quitting a program.

I am preparing a PR with a proposal correction.

dogfoodsilo4:
I've just stumbled into this same issue.  

I can confirm that the workaround works: -

```
 // Create
        midiIn = new MidiIn(index);
        midiIn.CreateSysexBuffers(20, 2);
        midiIn.SysexMessageReceived += In_Sysex;
        midiIn.Start();

        // Dispose
[...]

PR #909, opened 2022-06-04
correction proposal for issue #908 (sysex event infinite callback loop)

flyingoverclouds:
Proposal of correction for Naudio.Midi issue #908 (sysex event infinite callback loop) : 
- implement a disposeIsRunning instance flag tp avoid re-add the sysex buffer when dispose occur
- not firing C# SysexMessageReceived event when a zero length sysex message arrived


markheath:
Sorry took me a while to get round to this, and thanks for your contribution

PR #914, opened 2022-06-28
Set asio buffer size

smaciullator:
On AsioOut.InitRecordAndPlayback i've added a new optional parameter to set the buffer size on the current asio driver instance: if it's not specified (value = -1) it will use the preferred buffer size, else it will check if the value is between min and max size supported and also if it respect granularity; if the value is not coherent with the driver granularity it will search the closest good value and use it.
I've added a public method on AsioOut to get the current buffer size and also added a public property to expose the preferred buffer size (i've founde them to be useful).

I've tried this with different drivers and hardware and it seems to work well.
I've used:
- integrated generic audio card with asio4all
- m-audio m-track plus II with it's own asio driver and also with asio4all
- focusrite scarlet with it's own asio driver and also with asio4all
- behringer u-phoria um2 with it's own asio driver (but now it no longer exists and they force to use asio4all) and also with asio4all
All of this different setups has been tested on 2 different windows computers with windows 10.
Unfortunately i don't have access to any other audio cards or computers at the moment.


markheath:
Thanks for this contribution, and apologies it dropped at a time when I wasn't available to look at it in more detail. it's a genuinely useful addition to NAudio. 

The good news is that the ability to set the ASIO buffer size is something NAudio 3 now supports out of the box via the new `AsioDevice` API, so I'm going to close this, but your proposal lines up exactly with what we ended up building:

- **Request a specific size:** the `Asio*Options` records take a `BufferSize` (`int?`) — e.g. `AsioPlaybackOptions.BufferSize`, plus the recording and duplex equivalents. Leave it `null` to use the driver's preferred size.
- **Query the valid range up front:** `AsioDevice.Capabilities` exposes `BufferMinSize`, `BufferMaxSize`, `BufferPreferredSize` and `BufferGranularity`, so you can pick a sensible value rather than guessing.
- **Read back what's actually in use:** `AsioDevice.FramesPerBuffer` (also surfaced on `AsioOut.FramesPerBuffer`).
- **Driver-initiated changes:** handled via a buffer-size-changed callback, so a driver asking to renegotiate is dealt with too.

The main behavioural difference from this PR is that an out-of-range request throws `ArgumentOutOfRangeException` rather than silently snapping to the nearest granularity-valid size - we went with a failing loudly approach.

Closing as superseded, but thanks again for the contribution and the thorough testing notes across all those interfaces.

PR #921, opened 2022-07-08
Minor fixes:

JimBayne:
I've been using these 2 fixes since release 1.8. Thought they might be of use.

Fix 1) WaveFileWriter: 
- Don't dispose of memory streams in the Finalizer. It deletes them (which defeats WaveFileWriter's purpose)

Fix 2) Mp3FileReader.ReadNextFrame: 
- Don't try to read from a closed stream.
- Was happening intermittently in the past (probably due to some sort of timing + multi tasking issue)

Many thanks for this wonderful library.

### DIFF
--- a/NAudio.Core/Wave/WaveOutputs/WaveFileWriter.cs
+++ b/NAudio.Core/Wave/WaveOutputs/WaveFileWriter.cs
@@ -384,8 +384,8 @@ namespace NAudio.Wave
                     finally
                     {
                         Type Mem = typeof(MemoryStream); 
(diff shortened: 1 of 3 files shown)


markheath:
Many thanks for this contribution, and your support for NAudio. Apologies that I never got round to looking at this - I got a very long way behind on NAudio notifactions! Given the large changes in NAudio 3 I am closing some old PRs now, and I believe most of these fixes have been incorporated. Feel free to raise an issue if there are still problems in the latest code that you would like to see addressed

PR #929, opened 2022-08-08
Audio reader support streams

pvoelker:
Centralize logic to convert an audio file extension into an audio file type into a class 'AudioFileExtensions' with test cases around it.

Enhance 'AudioFileReader' class to allow a stream to be passed into a constructor.  Previously the class could only be constructed using a file name.


Alex-van-der-Peet:
Please merge this

pvoelker:
I’m still interested in seeing this get into the library to support stream based data sources

markheath:
Many thanks for this contribution - it has now been implemented (slightly differently) in NAudio 3

PR #943, opened 2022-09-10
Mark DirectSoundOut as Windows only

ds5678:
`DirectSoundOut` uses COM Interop and native method calls to `dsound.dll` and `user32.dll`. This pull request marks the class as Windows-only to make it more clear to users that this will not function on other platforms.

This pull request also adds a nontransitive reference to [PolySharp](https://github.com/Sergio0694/PolySharp), which is required to use the `SupportedOSPlatform` attribute on .NET Standard.

### DIFF
--- a/NAudio.Core/NAudio.Core.csproj
+++ b/NAudio.Core/NAudio.Core.csproj
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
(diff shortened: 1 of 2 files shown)


ds5678:
I rebased and updated my pull request to be more inline with my current coding standards.

markheath:
Thanks for this contribution. DirectSoundOut has now moved into NAudio.Dmo for NAudio 3, which targets net9.0-windows, so the whole assembly is platform-scoped

PR #947, opened 2022-09-27
Adds intrinsic acceleration to the `SilenceProvider`.

Zintom:
If the target framework supports intrinsics (.NET Core 3.0 and up), I define a new `Read` method which uses the power of Avx and Sse to clear the buffer.

Performance tests using BenchmarkDotNet indicate an 8% speedup using this method:

Better Silence Provider

I also improved documentation on the `Read` method.

If this PR is accepted I am more than happy to work on accelerated versions of other Wave providers

Any thoughts/ideas are more than welcome :)

### DIFF
--- a/NAudio.Core/NAudio.Core.csproj
+++ b/NAudio.Core/NAudio.Core.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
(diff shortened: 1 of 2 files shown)


markheath:
Thanks for this contribution, although with the new Span-based implementation of SilenceProvider in NAudio 3 is already vectorized by the runtime

PR #963, opened 2022-11-16
Add Stream constructor to CachedSound

jcummings2:
This is an update to Pull Request #602 in CacheSound's new location in NAudio.Extras. It consolidates common code into an `Init()` method.

I made the `Init()` method and property setters `protected` to allow for subclasses to more easily. Likewise, I added a protected parameterless constructor to support subclassing.


markheath:
Thanks for this contribution, it has now been implemented in NAudio 3

PR #982, opened 2022-12-12
Contribute ChannelMixerSampleProvider - arbitrary mixing of channels in a stream

antiduh:
This pull request is to contribute a new class, ChannelMixerSampleProvider, which allows a caller to remix the channels of a stream using an arbitrary matrix, with support for changing the number of output channels based on the configuration of the matrix. Example matrixes are provided in the ChannelMixMatrix class. For example, the mixer permits a caller to provide a 2-row/6-column matrix that would convert 2-channel input to 6-channel output.

I've tested manually using three audio devices - built-in motherboard realtek interface, a Motu M4, and a SoundBlaster G5.

### DIFF
--- a/NAudio.Core/Wave/SampleProviders/ChannelMixMatrix.cs
+++ b/NAudio.Core/Wave/SampleProviders/ChannelMixMatrix.cs
@@ -1,0 +1,72 @@
+﻿using System;
+
+namespace NAudio.Core.Wave.SampleProviders
+{
+    /// <summary>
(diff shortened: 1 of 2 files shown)


markheath:
Sorry for the extremely long delay - I got a long way behind on NAudio issues and PRs. But this is a nice feature - am gonna re-implement it as NAudio main branch has moved on a bit, but will make sure you get credited for this contribution.

markheath:
Closing as reimplemented against main. Thanks for this contribution

PR #1004, opened 2023-02-10
WaveFormRenderer.Render() parameter changed.

DjNemas:
The renderer use the `WaveStream` class now. I updated the guide to correct implementation, with the newer `AudioFileReader : WaveStream` class.


nightblade9:

PR #1009, opened 2023-03-02
Correction in PlaybackLatency description

MAtt5816:


markheath:
Sorry for the delay getting back to you on this. It was actually quite hard to find any official documentation on what the units of this parameter are, but I agree it does seem to be samples (or perhaps more accurately "sample frames" - i.e. it hasn't been multiplied by the channel count)

PR #1010, opened 2023-03-05
Add support for capture options in WASAPI loopback

jaimeadf:
This adds the ability to set options of `WasapiCapture` when constructing `WasapiLoopbackCapture`. I've tested it by creating an application that keeps track of the number of bytes received in `DataAvailable` and measures the time between each event.

### DIFF
--- a/NAudio.Wasapi/WasapiLoopbackCapture.cs
+++ b/NAudio.Wasapi/WasapiLoopbackCapture.cs
@@ -28,6 +28,27 @@ namespace NAudio.Wave
         }
 
         /// <summary>
+        /// Initializes a new instance of the <see cref="WasapiLoopbackCapture"/> class.
+        /// </summary>
+        /// <param name="captureDevice">The capture device.</param>
+        /// <param name="useEventSync">true if sync is done with event. false use sleep.</param>
+        public WasapiLoopbackCapture(MMDevice captureDevice, bool useEventSync) :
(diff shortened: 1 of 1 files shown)


nightblade9:

markheath:
Thanks for this contribution and apologies it took me so long to get round to looking at it properly. These are valuable features but the way we will deliver these in NAudio 3 is going to be slightly different.


The two options you're exposing here - event vs polling sync, and the capture buffer length - are both first-class in NAudio 3's new `WasapiRecorder`, which replaces `WasapiCapture` / `WasapiLoopbackCapture` (both are now `[Obsolete]`). The equivalent of what this PR enables is:

```csharp
using var recorder = new WasapiRecorderBuilder()
    .WithLoopbackCapture()      // system loopback (or WithProcessLoopback(pid) for per-process)
    .WithPollingSync()          // or WithEventSync() (the default) — your useEventSync flag
    .WithBufferLength(50)       // your audioBufferMillisecondsLength (default 100ms)
    .Build();

recorder.DataAvailable += (s, a) => { /* ... */ };
recorder.StartRecording();
```

The builder combines these freely with loopback and adds a lot more on top (format request, MMCSS thread priority, IAudioClient3 low-latency, raw mode, per-process loopback, default-device stream routing, and AEC), so on `main` there's nothing further to add — your scenario is already covered.

Thanks again, and hopefully the new features in NAudio 3 meet your needs

PR #1016, opened 2023-03-30
Implemented classes for missing interfaces needed to navigate device topology

gribunin:
DeviceTopology class (available via property of MMDevice) has properties to get connectors information (Connector class), but the implementation of interfaces like IPart and other related interfaces is missing. This pull request adds the missing classes and adds a demo which navigates device topology of the selected MMDevice.


markheath:
Sorry for the delay, and thanks for the contribution, looks really useful

PR #1017, opened 2023-03-31
fixed WASAPI helper threads not being background

lostmsu:
This caused the app that uses WasapiCapture and/or WasapiOut to not exit unless all instances are disposed.


markheath:
thanks, I wanted to check exactly what the implications are first but according to [this doc](https://learn.microsoft.com/en-us/dotnet/standard/threading/foreground-and-background-threads) "Background threads are identical to foreground threads with one exception: a background thread does not keep the managed execution environment running.". So that sounds fine to me

PR #1032, opened 2023-05-27
Fix floating point time conversion, move to integer with 3 digits in milliseconds

Quake4:


nightblade9:

markheath:
Thanks for this contribution but we have resolved this issue in a different way (see #754)

PR #1048, opened 2023-07-21
Need access to writer to override class to add additional chunks

copsonit:
I needed to add the smpl chunk to support loops when writing WAV. If anybody see how to extend the WaveFileWriter to support additional chunks without access to the private writer, please let me know!

### DIFF
--- a/NAudio.Core/Wave/WaveOutputs/WaveFileWriter.cs
+++ b/NAudio.Core/Wave/WaveOutputs/WaveFileWriter.cs
@@ -140,6 +140,11 @@ namespace NAudio.Wave
         }
 
         /// <summary>
+        /// The wave writer
+        /// </summary>
+        public BinaryWriter Writer => writer;
+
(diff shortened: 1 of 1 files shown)


markheath:
Have you looked at `CueWaveFileWriter` which shows inheriting `WaveFileWriter` allowing additional chunks to be written.

copsonit:
I must have been too influenced how the "fact" chunk was implemented and tried to write the "smpl" chunks before the "data", but of course there should be no problems of appending the "smpl" chunk at the end instead, like how the CueWaveFileWriter does, which I completely missed. Thanks for the tip! So no need for exposing writer so I am closing this PR.

markheath:
Great, glad that sorted your issue. I'd prefer not to expose the writer, but maybe in future some kind of generic chunk writer capability should be added to `WaveFileWriter` to avoid having to specialize for every chunk type

PR #1113, opened 2024-01-23
Changed error details for MmException

SMSapereAude:
I think current error description is vague, as all it tells is i.e. `NoDriver calling waveOutSetVolume`. The point of this pull request is to make debugging and troubleshooting easier. (especially for non-experienced library users)


SMSapereAude:
@markheath could you please take a look?

markheath:
thanks for this contribution, but if we are to make a change here, it will more likely be the approach from #1192

Issue #819, opened 2021-08-03
WASAPI built-in resampler: Add AUDCLNT_STREAMFLAGS_AUTOCONVERTPCM flag


feoff3:
I discovered that there is a built-in resampler in WASAPI that allows to use any wave format with WASAPI.
It is turned on by AUDCLNT_STREAMFLAGS_AUTOCONVERTPCM flag , see
https://docs.microsoft.com/en-us/windows/win32/coreaudio/audclnt-streamflags-xxx-constants
The flag is succesfully used in miniaudio and portaudio C++ open source libraries. 

Adding this option to NAudio WASAPI implementation would greatly simplify the life of Windows-based developers as manual resampling is cumbersome and sometimes results in pretty much unpredictable results. Use of something built-in instead looks like a more reliable and simple option.

I was wondering if it works in NAudio. I was able to change the NAudio code inserting the flag into NAudio lib+disabling IsFormatSupported() call for NAudio.Wasapi/WasapiCapture.cs. 
This allowed me to specify any format for WasapiCapture even if it is not supported by the endpoint device. 

For this, I changed  AudioClientStreamFlags() code:
```
        protected AudioClientStreamFlags GetAudioClientStreamFlags()
        {
            // autoresampling flags. works starting from Win8
            uint AUDCLNT_STREAMFLAGS_AUTOCONVERTPCM = 0x80000000;
            uint AUDCLNT_STREAMFLAGS_SRC_DEFAULT_QUALITY = 0x08000000;
            uint streamFlags = AUDCLNT_STREAMFLAGS_AUTOCONVERTPCM | AUDCLNT_STREAMFLAGS_SRC_DEFAULT_QUALITY;
            return (AudioClientStreamFlags)streamFlags;
        }
```

and deleted IsFormatSupported() check from https://github.com/naudio/NAudio/blob/fb35ce8367f30b8bc5ea84e7d2529e172cf4c381/NAudio.Wasapi/WasapiCapture.cs#L127

Let me know how to proceed: propose a patch or let you incorporate this flag in a way you prefer (e.g. having a separate class, or  whatever that suits NAudio code design)

markheath:
Wow, didn't know that was in there, and would be great to get in there (especially if this works for playback too)

In terms of fitting with NAudio design - a couple of thoughts
1. I'd probably make auto-resampling optional but on by default. e.g. a class property that you can change before beginning capture
2. I still think you need to check if the format is supported since the user could have asked for something unusual like MP3 format

[...]

PR #1122, opened 2024-02-10
Fix a bug that WasapiCapture could not use exclusive mode

tt3333:
When using exclusive mode with WasapiCapture, the InitializeCaptureDevice throws an exception.
In #819 it says "you can't use it with WASAPI exclusive mode playback", but it seems that auto convert cannot be used in exclusive mode for recording as well as playback.

```
System.ArgumentException: Value does not fall within the expected range.
   at System.Runtime.InteropServices.Marshal.ThrowExceptionForHR(Int32 errorCode)
   at NAudio.CoreAudioApi.AudioClient.Initialize(AudioClientShareMode shareMode, AudioClientStreamFlags streamFlags, Int64 bufferDuration, Int64 periodicity, WaveFormat waveFormat, Guid audioSessionGuid)
   at NAudio.CoreAudioApi.WasapiCapture.InitializeCaptureDevice()
   at NAudio.CoreAudioApi.WasapiCapture.StartRecording()
```

### DIFF
--- a/NAudio.Wasapi/WasapiCapture.cs
+++ b/NAudio.Wasapi/WasapiCapture.cs
@@ -124,6 +124,11 @@ namespace NAudio.CoreAudioApi
 
             long requestedDuration = ReftimesPerMillisec * audioBufferMillisecondsLength;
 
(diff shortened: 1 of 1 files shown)


nightblade9:

PR #1125, opened 2024-02-14
WaveOutEvent device capabilities

davidvesely:
Adding `DeviceCount` and `GetCapabilities` to `WaveOutEvent`, unifying it with `WaveInEvent`.

### DIFF
--- a/NAudio.WinMM/WaveOutEvent.cs
+++ b/NAudio.WinMM/WaveOutEvent.cs
@@ -65,6 +65,24 @@ namespace NAudio.Wave
         }
 
         /// <summary>
+        /// Returns the number of Wave Out devices available in the system
+        /// </summary>
+        public static int DeviceCount => WaveInterop.waveOutGetNumDevs();
+
+        /// <summary>
+        /// Retrieves the capabilities of a waveOut device
+        /// </summary>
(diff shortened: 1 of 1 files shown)


nightblade9:

markheath:
Thanks for this, @davidvesely, and sorry for the long delay in responding.

The good news is that NAudio 3 (`main`) already does exactly what this PR sets out to do. As part of the NAudio 3 cleanup, `WaveOutEvent` was renamed to `WaveOut` (`WaveOutEvent` now lives on only as an `[Obsolete]` subclass shim), and the renamed `WaveOut` already exposes the unified API you proposed — essentially line-for-line the same code:

```csharp
public static int DeviceCount => WaveInterop.waveOutGetNumDevs();
public static WaveOutCapabilities GetCapabilities(int devNumber) { ... }
```

`WaveIn` (formerly `WaveInEvent`) has the matching pair, so input/output are now unified. 

Given that, I'm going to close this PR as already-resolved on `main`. Really appreciate the contribution — it was the right idea, but the rename work has superseded it.

XeonG:
Just seen this and this should go in v2, main is a newer .net version that is simply not compatible with current Unity c# net versions yet.. maybe in a another year it would be, but currently not.

PR #1153, opened 2024-04-30
IWavePlayer and IWaveIn cross-platform implementations via SDL2

alextnull:
IWavePlayer and IWaveIn cross-platform implementations via SDL2.
This code is based on `NAudio`, `SDL`, `SIPSorceryMedia.SDL2`, `SDL2-CS`, `osu-framework`, and other `SDL` related projects.
Don't expect it to be stable, tested and work the way you assume it will.

```csharp
// Recording
IWaveIn waveIn = new WaveInSdl();
waveIn.DataAvailable += (s, e) =>
{
    // ...
};
waveIn.StartRecording();

// Playback
IWavePlayer waveOut = new WaveOutSdl();
IWaveProvider waveReader = new WaveFileReader("Example.wav");
waveOut.Init(waveReader);
waveOut.Play();
```


markheath:
Thanks for this extensive contribution. In its current state it wouldn't be able to come into NAudio unfortunately, although maybe in the future an SDL binding would be a nice feature (maybe SDL3). The Avalonia demo project would also bring a major dependency into the build system which I'm not sure I want to take on at this time.

PR #1207, opened 2025-02-06
Implement GCHandle over AllocHGlobal in Wasapi AudioClient

JustArion:
The original implementation asked for an alternative to using `AllocHGlobal` since we can't use the `fixed` keyword nor use the `Unsafe` class as it isn't implemented in .Net Standard 2.0

Minor House Keeping: Discarded unused variable & removed redundant type cast.

#### ! Implementation of SetPropertiesForClient is `untested`

### DIFF
--- a/NAudio.Wasapi/CoreAudioApi/AudioClient.cs
+++ b/NAudio.Wasapi/CoreAudioApi/AudioClient.cs
@@ -20,6 +20,20 @@ namespace NAudio.CoreAudioApi
         private AudioStreamVolume audioStreamVolume;
         private AudioClientShareMode shareMode;
 
+
+        private static void SetPropertiesForClient(IAudioClient2 client, in AudioClientProperties properties)
+        {
+            var handle = GCHandle.Alloc(properties, GCHandleType.Pinned);
+
+            try
+            {
+                client.SetClientProperties(handle.AddrOfPinnedObject());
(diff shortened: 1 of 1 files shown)


markheath:
Thanks for this, @JustArion, and for the write-up of the AllocHGlobal problem. I've been using some AI to help me triage the backlog of PRs and issues, and just got round to this.

This PR targets [the NAudio 2 branch](https://github.com/naudio/NAudio/blob/claude/review-pr-1207-6vSfr/release/2.x), and its motivation - working around the lack of fixed/Unsafe in .NET Standard 2.0 - no longer applies on main (NAudio 3): NAudio.Wasapi now targets net9.0-windows only, and AudioClient.ActivateAsync has since been rewritten for source-generated COM marshalling plus a multi-vtable CCW fix, so most of these changes conflict with or are obsoleted by that work.

On the approach itself: GCHandle.Alloc(..., Pinned) only works here because AudioClientProperties happens to be fully blittable today — it boxes the struct (so it still allocates) and would throw at runtime if a non-blittable field were ever added, whereas Marshal.StructureToPtr stays correct. On modern .NET the better direction is allocation-free pointer passing rather than GCHandle pinning.

Hopefully what's in NAudio 3 meets your needs, but let me know if there's anything else we're still missing.

Issue #1205, opened 2025-02-06
MMDevice PropertyStore Performance Concerns


JustArion:
With the following reproducible example we show that there's significant overhead in accessing properties that use the PropertyStore.
The example includes the overhead of writing to the Console. but in this case that's negligible. 
```cs
using System.Diagnostics;
using NAudio.CoreAudioApi;

using var enumerator = new MMDeviceEnumerator();
using var device = enumerator.GetDefaultAudioEndpoint(DataFlow.Render, Role.Multimedia);

var ts = Stopwatch.GetTimestamp();
Console.WriteLine(device.DeviceFriendlyName);
Console.WriteLine($"Operation 1 took {Stopwatch.GetElapsedTime(ts).TotalMilliseconds}ms");

ts = Stopwatch.GetTimestamp();
Console.WriteLine(device.DeviceFriendlyName);
Console.WriteLine($"Operation 2 took {Stopwatch.GetElapsedTime(ts).TotalMilliseconds}ms");

ts = Stopwatch.GetTimestamp();
Console.WriteLine(device.DeviceFriendlyName);
Console.WriteLine($"Operation 3 took {Stopwatch.GetElapsedTime(ts).TotalMilliseconds}ms");
```
```
CORSAIR HS80 RGB Wireless Gaming Headset
Operation 1 took 244.9605ms
CORSAIR HS80 RGB Wireless Gaming Headset
Operation 2 took 237.037ms
CORSAIR HS80 RGB Wireless Gaming Headset
Operation 3 took 232.7609ms
```

Lets see what's going on.
In [MMDevice.cs#L202](https://github.com/naudio/NAudio/blob/master/NAudio.Wasapi/CoreAudioApi/MMDevice.cs#L202) is the property `DeviceFriendlyName` we have 3 invocations of note. A method call to `GetPropertyInformation` (Negligable), a method call to `propertyStore.Contains` (130ms overhead), and then finally the "array" accessor `this[PropertyKey]` (130ms overhead).

@RequiDev has a proposal on how to lower this overhead to pretty much nothing (0.2ms)
In [PropertyStore.cs](https://github.com/naudio/NAudio/blob/master/NAudio.Wasapi/CoreAudioApi/PropertyStore.cs#L68-L101)
2 members are changed to:
```cs
        public bool Contains(PropertyKey key)
        {
            var result = storeInterface.GetValue(ref key, out var propVariant);
[...]

PR #1208, opened 2025-02-06
Wasapi PropertyStore - Significantly improve performance on accessing properties

JustArion:
Addresses #1205.

- 650-1250x improvement (130 - 250ms -> 0.2ms) on accessing properties from the [PropertyStore](https://github.com/naudio/NAudio/blob/master/NAudio.Wasapi/CoreAudioApi/PropertyStore.c) which is used by [MMDevice](https://github.com/naudio/NAudio/blob/master/NAudio.Wasapi/CoreAudioApi/MMDevice.cs)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,10 @@ _ReSharper*/
 *.ncrunchproject
 *.orig
 *.lock.json
+# Jetbrains
+*.idea
 .vs/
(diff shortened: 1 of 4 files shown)


markheath:
thanks, looks a nice improvement

Issue #878, opened 2022-02-17
Capture specific process with WasapiLoopbackCapture


joseluisct:
Hello,

I would like to use NAudio WasapiLoopbackCapture with the new possibility of capture loopback audio from (or excluding) a specific process. This requires Win 10 20348 or later.

There is a new initialization structure of the ActivateAudioInterfaceAsync Win32 API that allows you to pass a process to be included or excluded (and it's children).

I have been playing with the sample code on the docs and it's working OK but I would like to use it from my .Net apps that use NAudio. It would be a very good opportunity for audio apps.
https://docs.microsoft.com/en-us/samples/microsoft/windows-classic-samples/applicationloopbackaudio-sample/

```            
AUDIOCLIENT_ACTIVATION_PARAMS audioclientActivationParams = {};
audioclientActivationParams.ActivationType = AUDIOCLIENT_ACTIVATION_TYPE_PROCESS_LOOPBACK;
       audioclientActivationParams.ProcessLoopbackParams.ProcessLoopbackMode = includeProcessTree ?
       PROCESS_LOOPBACK_MODE_INCLUDE_TARGET_PROCESS_TREE : PROCESS_LOOPBACK_MODE_EXCLUDE_TARGET_PROCESS_TREE;
audioclientActivationParams.ProcessLoopbackParams.TargetProcessId = processId;
PROPVARIANT activateParams = {};
activateParams.vt = VT_BLOB;
activateParams.blob.cbSize = sizeof(audioclientActivationParams);
activateParams.blob.pBlobData = (BYTE*)&audioclientActivationParams;
wil::com_ptr_nothrow<IActivateAudioInterfaceAsyncOperation> asyncOp;
RETURN_IF_FAILED(ActivateAudioInterfaceAsync(VIRTUAL_AUDIO_DEVICE_PROCESS_LOOPBACK, __uuidof(IAudioClient), 
        &activateParams, this, &asyncOp));
// Wait for activation completion
m_hActivateCompleted.wait();
```

Thanks.

markheath:
This is really cool. I don't have much time for working on NAudio these days but I'd love to get this in. Will have to find some time for experimenting with it

markheath:
I gave this a good try, but something isn't quite working. I think I'm close, but the IAudioClient I end up creating fails with weird COM errors. Here's what I tried: https://github.com/naudio/NAudio/commit/446096a0495950342a3594bb57bea00bfdbad44e

TomieAi:
[...]

PR #1225, opened 2025-07-01
Implement Wasapi Process Specific Audio Capture

JustArion:
This work follows off-of @markheath's [process-audio-capture](https://github.com/naudio/NAudio/tree/process-audio-capture) branch and so, the PR **targets the same branch** for merging instead of master.

Modified files:

- #### `NAudio.Wasapi\CoreAudioApi\AudioClient.cs`
During tests for process specific audio capture, I stumbled on BufferSize returning invalid values sometimes. The change allows you to set a breakpoint to monitor for these invalid values. i.e negative buffer sizes or 0 sized buffer sizes. This is an issue that presents itself in the current implementation of it but is mitigated by detecting this and setting a fallback size buffer. We'll continue on this issue in the next modified file excerpt. 

MixFormat bugfix:
The `HRESULT` check prevents an access violation if the MixFormat is not supported.
0x80004001 (`E_NOTIMPL`) is larger than an int  and overflows into the negative (`-2147467263`). We check for `-0x7FFFBFFF` which is simply the hex version of `-0x7FFFBFFF` as an int.

- #### `NAudio.Wasapi\WasapiCapture.cs`

The `waveFormat` is unsupported in Wasapi Process Capture. We simply remove it from the root constructor and put it into the closest reasonable constructor that does need it (Wasapi Capture uses the root constructor that other public constructors forward into).

The fallback buffer implementation: During testing, very rarely, I would get a negative or zero buffer size for Wasapi Capture. I believe this could have been a bug in certain Windows builds as I haven't experienced it on build 26100 going forward. If I remember correctly the largest observed size the buffer needed was somewhere upwards of 10k bytes. The fallback buffer is considerably larger (around 2.4mb). This mitigation prevents process specific audio capture from simply playing back nothing and saving nothing when recording process audio. The reason for such a large fallback buffer is because the type for `audioClient.BufferSize` is int while the underlying COM implementation is a uint. Any negative buffer sizes can be accounted as a really large buffer size (bigger than int.MaxValue) This is still believed to be a bug on certain Windows builds and would certainly break C# builds if we use uint for array sizes since we can't even do `new byte[int.MaxValue]` let alone something bigger than that.
Sticking with int is a good call here and just to use a fallback buffer.

Lastly, we change from using AllocHGlobal in `public static async Task<WasapiCapture> CreateForProcessCaptureAsync(int processId, bool includeProcessTree)` to using GCHandle.Alloc instead. We also use the GUID attached to the IAudioClient interface as we don't need to re-create GUIDS every single time.

```diff
- var IID_IAudioClient = new Guid("1CB9AD4C-DBFA-4c32-B178-C2F568A703B2");
- NativeMethods.ActivateAudioInterfaceAsync(VIRTUAL_AUDIO_DEVICE_PROCESS_LOOPBACK, IID_IAudioClient, propVariant, icbh, out var activationOperation);
+ NativeMethods.ActivateAudioInterfaceAsync(VIRTUAL_AUDIO_DEVICE_PROCESS_LOOPBACK, typeof(IAudioClient).GUID, hActivateParams.AddrOfPinnedObject(), icbh, out var activationOperation);
```

Addresses #878 in which we go into more detail on the reasoning behind much of what this PR addresses.

---

### DIFF
--- a/NAudio.Wasapi/CoreAudioApi/AudioClient.cs
+++ b/NAudio.Wasapi/CoreAudioApi/AudioClient.cs
@@ -107,8 +107,10 @@ namespace NAudio.CoreAudioApi
         {
             get
             {
-                Marshal.ThrowExceptionForHR(audioClientInterface.GetBufferSize(out uint bufferSize));
-                return (int) bufferSize;
+                var result = audioClientInterface.GetBufferSize(out var bufferSize);
+                Marshal.ThrowExceptionForHR(result);
+                var retVal = (int)bufferSize; // Allows for setting a breakpoint on an invalid value ( < 1 )
+                return retVal;
             }
         }
 
@@ -267,6 +269,15 @@ namespace NAudio.CoreAudioApi
             IntPtr pointerToPtr = GetPointerToPointer(); // IntPtr.Zero; // Marshal.AllocHGlobal(Marshal.SizeOf<WaveFormatExtensible>());
             closestMatchFormat = null;
             int hresult = audioClientInterface.IsFormatSupported(shareMode, desiredFormat, pointerToPtr);
+
+            // -2147467263 0x80004001
+            // E_NOTIMPL
+            // Since the HRESULT isn't checked until after we read the pointer previously
+            // the pointer is uninitialized memory and dereferencing it throws an AV exception
+            if (hresult == -0x7FFFBFFF)
+            {
+                return false;
+            }
 
             var closestMatchPtr = Marshal.PtrToStructure<IntPtr>(pointerToPtr);
 
--- a/NAudio.Wasapi/WasapiCapture.cs
+++ b/NAudio.Wasapi/WasapiCapture.cs
@@ -18,6 +18,7 @@ namespace NAudio.CoreAudioApi
     {
         private const long ReftimesPerSec = 10000000;
         private const long ReftimesPerMillisec = 10000;
+        private const int FALLBACK_BUFFER_LENGTH = 10000;
         private volatile CaptureState captureState;
         private byte[] recordBuffer;
         private Thread captureThread;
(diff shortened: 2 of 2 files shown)


JustArion:
A simple proof of concept Console Application is as follows:
```cs
using System.Diagnostics;
using NAudio.CoreAudioApi;
using NAudio.Wave;

Console.WriteLine("Enter Process Id");
var pid = int.Parse(Console.ReadLine()!);

var cap = await WasapiCapture.CreateForProcessCaptureAsync(pid, true); // Spotify

var buffer = new BufferedWaveProvider(cap.WaveFormat)
{
    ReadFully = false,
    BufferDuration = TimeSpan.FromSeconds(15)
};

cap.DataAvailable += (_, data) => buffer.AddSamples(data.Buffer, 0, data.BytesRecorded);
Console.WriteLine($"[*] Recording for {buffer.BufferDuration.TotalSeconds} sec");
cap.StartRecording();

await Task.Delay(buffer.BufferDuration);
cap.StopRecording();

using var recording = new WasapiOut();
var tcs = new TaskCompletionSource();

recording.Init(buffer);
Console.WriteLine("[*] Playing from buffer");
Console.WriteLine($"[*] Buffer size is {buffer.BufferedBytes} bytes ({BytesToMB(buffer.BufferedBytes):F}mb)");
var ts = Stopwatch.GetTimestamp();
recording.Play();
recording.PlaybackStopped += (_, _) => tcs.SetResult();
    
await tcs.Task;
Console.WriteLine($"[*] Playback took {Stopwatch.GetElapsedTime(ts)}");
return;

float BytesToMB(int bytes) => bytes / 1024f / 1024;
```

JustArion (commented):
The implementation was discussed further [here](https://github.com/naudio/NAudio/issues/878#issuecomment-2648003055)

markheath:
well done for progressing this - has been a long time but remember getting stuck on it. not sure when I'll get a chance to test it myself, but will bring it into the process-audio-capture branch and then can see what it would take to get it into the master branch

PR #1234, opened 2025-09-13
Improve multi-channel playback compatibility in WASAPI exclusive mode

kashiken:
This PR improves playback compatibility for multi-channel audio devices in WASAPI exclusive mode.

## Purpose of Change
In the current code, dwChannelMask in the WaveFormatExtensible constructor is set by simply shifting and OR-ing bits based on the number of channels.  
However, actual multi-channel audio devices return values that differ from this, because Windows expects the standard speaker placements defined in ksmedia.h and devices follow that specification.

## Changes
* Add an optional channelMask argument to the WaveFormatExtensible constructor.
  * If channelMask is not zero, dwChannelMask will be set to channelMask.
  * Otherwise, dwChannelMask is set as before.
* Extend GetFallbackFormat() to try additional channel masks as well as sampling rates, channel counts, and bit depths.
  * The candidates are taken from ksmedia.h in the Windows Driver Kit 10.0.19041.0.
  * The related documentation can be found here: https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ksmedia/ns-ksmedia-waveformatextensible

## Tests
Tested with a 4-channel USB audio device (channel mask 0x33 = KSAUDIO_SPEAKER_QUAD). 
It cannot play back in WASAPI exclusive mode with the current code, but playback works after this change.

### DIFF
--- a/NAudio.Wasapi/WasapiOut.cs
+++ b/NAudio.Wasapi/WasapiOut.cs
@@ -269,7 +269,7 @@ namespace NAudio.Wave
             if (!bitDepthsToTry.Contains(24)) bitDepthsToTry.Add(24);
             if (!bitDepthsToTry.Contains(16)) bitDepthsToTry.Add(16);
 
-            var channelMaskToTry = new List<int>() { 0 };
+            var channelMasksToTry = new List<int>() { 0 };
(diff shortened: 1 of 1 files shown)


markheath:
looks useful. Did you also check that this still works on stereo devices?

Issue #203, opened 2017-06-17
AcmMp3FrameDecompressor uses wrong bitrate


PixelSnake:
I am streaming MP3 files from a server via `BufferedWaveProvider` and `AcmMp3FrameDecompressor` exactly like in the NAudio demo application. For most files, this works just well, but there are some files, which are encoded in 44,1kHz, just like all the others, which get decoded in a different bitrate (i observed 48kHz and 32kHz). This only happens for these exact files and always in the same way (not a random bitrate but always the same, wrong, bitrate)

```
frame = Mp3Frame.LoadFromStream(readFullyStream);
// frame.SampleRate is wrong
```

markheath:
Sometimes corruption (or maybe album art) at the start of an MP3 file can cause NAudio to wrongly infer the sample rate of the file. There is a long-standing need for us to wait for a larger number of consecutive frames with the same sample rate before deciding that we've determined the sample rate.

realares:
NAudio totally ignores the existing ID3 Tags. First find out where these TAGs end and the first audio frame begins before evaluate the frame header. Quick and dirty, Skip at least half of the file.

markheath:
Closing as fixed. The root cause here — false frame detections from album art / ID3 padding causing the wrong sample rate to be inferred — has been addressed by requiring two consecutive valid frame headers with matching MPEG version, layer, and sample rate before accepting a candidate header.

Relevant changes:
- `ffda548` — added `IsLikelyAudioFrameSequence` multi-frame confirmation in `Mp3Frame.LoadFromStream` (the fix I described in my 2017 comment above).
- `24c796c` (#1253) — fixed an edge case where the new check was over-rejecting valid frames near end-of-file (e.g. when an ID3v1 trailer follows the last audio frame).

For non-seekable streams (`BufferedWaveProvider` + `AcmMp3FrameDecompressor` style streaming, as in the original report) the multi-frame check is skipped to preserve streaming behaviour, but for seekable sources MP3 parsing is now significantly more robust.

As a side note, on modern Windows we generally recommend `MediaFoundationReader` over `Mp3FileReader` / `AcmMp3FrameDecompressor` for MP3 playback — it's more robust against malformed files and doesn't depend on the legacy ACM codec. Thanks for the report!

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #1172, opened 2024-07-20
Got a frame at sample rate 32000, in an MP3 with sample rate 12000. Mp3FileReader does not support sample rate changes.'


satpowers:
I am trying to load file to memorystream then play it, but I get error:
Got a frame at sample rate 32000, in an MP3 with sample rate 12000. Mp3FileReader does not support sample rate changes.'
on the line: var mp = new Mp3FileReader(ms);

this is my code
                byte[] bytes = File.ReadAllBytes(testFile);
                int i = bytes[bytes.Length - 1];
                var ms = new MemoryStream(bytes, 0, bytes.Length, false, true);
                var mp = new Mp3FileReader(ms);

markheath:
Closing this as we have made several improvements to MP3 parsing that should eliminate this issue
In addition, on Windows I recommend using MediaFoundationReader for MP3s, and for cross-platform, use SoundFilePlayer (uses libsndfile)

Issue #763, opened 2021-04-07
[BUG] Artwork in ID3 mp3 tags causing sample rate error.


RSDNTWK:
Thanks to user bgr at Altfuture we have confirmed that this issue https://github.com/naudio/NAudio/issues/229 is caused by artwork being present in the ID3tag data of mp3's causing sample rate errors.

Milkitic:
Old issue but if you just want to play audio, you can try to remove tags before reading.
My temporary solution:
https://github.com/Milkitic/Milki.Extensions/blob/main/src/Milki.Extensions.MixPlayer/NAudioExtensions/Wave/SmartWaveReader.cs#L145-L166

markheath:
I recommend using `MediaFoundationReader` instead of `Mp3FileReader` if possible, as that will not have this issue with ID3 tags

markheath:
No `MediaFoundationReader` does not use ACM, but it does use Media Foundation which is part of Windows.
To do MP3 decoding you can hook NLayer into NAudio, but you can't do AAC.

markheath:
Closing this as we've fixed a bunch of MP3 parsing errors in NAudio 3
Also, on Windows I recommend using MediaFoundationReader for playing MP3, and there is also now SoundFileReader for cross platform MP3 playback (via libsndfile)

Issue #583, opened 2020-01-25
Cannot open an MP3 file


mc0re:
Error message:
Got a frame at sample rate 11025, in an MP3 with sample rate 44100. Mp3FileReader does not support sample rate changes.

I'm not sure how I can attach the file...

markheath:
Often this means a corrupt MP3 file (maybe album art being interpreted as MP3). If you can use MediaFoundationReader instead, that may work better.

mc0re:
I should've mentioned, the file is played fine by MPC-HC.

mc0re:
Here is the file: https://mega.nz/#!4tAzBI5Z!pVWQWe32REKSKQsDwE92HYxtuuCGV8zYT2VTksPv3wE

Disclaimer: I am not distributing anything illegally, this file is shared for the purpose of software debugging.

markheath:
Closing this as we've fixed MP3 frame parsing errors in the latest NAudio 3
You can also use MediaFoundationReader to read MP3s (recommended on Windows), or SoundFileReader (recommended for cross platform)

PR #1253, opened 2026-03-20
Fix false MP3 sample rate change errors near end of file

### LINKED ISSUE #229: NAudio does not support VBR(Got a frame at sample rate 16000, in an MP3 with sample rate 44100. Mp3FileReader does not support sample rate changes)
My requirement is "change the channel as 1, bitrate as 32kbps and splite the time span the mp3 file as 2 minuts". 
```
 public static Stream ProcessAudioStream(Stream inputStream)
        {
            Stream fs = new MemoryStream();
            Stream outputStream = new MemoryStream();           
            try
            {                

                Stream tempStream = new MemoryStream(); 

                using (var waveStream = new Mp3FileReader(inputStream))
                using (WaveFileWriter waveFileWriter = new WaveFileWriter(tempStream, new Mp3WaveFormat(44100,1,8,32000)))
                {
                    byte[] bytes = new byte[waveStream.Length];
                    waveStream.Read(bytes, 0, (int)waveStream.Length);
                    waveFileWriter.Write(bytes, 0, bytes.Length);
                    waveFileWriter.Flush();
                    fs = BytesToStream(bytes);
                }
                fs.Position = 0;

                using (var mp3 = new Mp3FileReader(fs))
                //Got a frame at sample rate 16000, in an MP3 with sample rate 44100. Mp3FileReader does not support sample rate changes
                {
                    Mp3Frame frame;
                    while ((frame = mp3.ReadNextFrame()) != null)
                    {
                        if ((int)mp3.CurrentTime.TotalSeconds <= 120)//2 mintues
                        {
                            `outputStream.Write(frame.RawData,` 0, frame.RawData.Length);
                        }
                    }
                }
           }
            catch (Exception e)
            {
                string msg = e.Message;

            }
            return outputStream;
        }

```
Mp3FileReader is throwing an exception: 
**Got a frame at sample rate 16000, in an MP3 with sample rate 44100. Mp3FileReader does not support sample rate changes.**

Would someone give me some suggestions for this error? thanks
In additions, you can click below url to download my source code. Use VS2015 to open the project, thanks.
 https://pan.baidu.com/s/1qYiEqle

My email is 54373175@qq.com 
Let me know if you have any comments for this error.

markheath:
## Summary
- Fixes a long-standing bug where `Mp3FileReader` throws "Mp3FileReader does not support sample rate changes" on valid MP3 files
- The `IsLikelyAudioFrameSequence` consecutive-frame validation (added in ffda5488) rejected valid frames near the end of the file when its look-ahead landed on non-audio data (e.g. an ID3v1 tag), causing the byte scanner to find a false frame header with a different sample rate
- The fix accepts the candidate frame when there isn't room for another full frame in the remaining stream, correctly distinguishing end-of-audio-data from a false positive mid-stream

## Root cause
For the last 1-2 frames in a file with an ID3v1 tag, `IsLikelyAudioFrameSequence` seeks forward by `FrameLength` to verify consecutive frames. When that lands on the 128-byte ID3v1 tag instead of another MP3 frame:
1. The valid frame is rejected
2. The byte scanner searches forward through compressed audio data
3. It finds random bytes (e.g. `FF FF 8B 88`) that parse as a valid header with a different sample rate (e.g. 32kHz Layer 1)
4. That false header's large calculated frame length extends past the end of the file, so the consecutive-frame check early-returns true
5. `ValidateFrameFormat` sees the sample rate mismatch and throws

Fixes #229, #292, #583, #763, #1172

## Test plan
- [x] Verified fix against the specific file that triggered the bug (CBR 192kbps 44.1kHz with ID3v1 tag)
- [x] Tested against 6 user-reported problem MP3 files — all pass (M2M.mp3 was actively failing before the fix)
- [x] 5 of 6 files now correctly read 2 additional frames at the end that were previously silently dropped
- [x] All 52 existing MP3 unit tests pass

### DIFF
--- a/NAudio.Core/FileFormats/Mp3/Mp3Frame.cs
+++ b/NAudio.Core/FileFormats/Mp3/Mp3Frame.cs
@@ -143,14 +143,17 @@ namespace NAudio.Wave
                     var nextFrame = new Mp3Frame();
                     if (!IsValidHeader(headerBytes, nextFrame))
                     {
-                        return false;
+                        // If there isn't enough room for another frame of similar size,
+                        // we're likely at the last audio frame and have hit non-audio data
+                        // (e.g. an ID3v1 tag at the end of the file)
+                        return nextFrameOffset + firstFrame.FrameLength >= input.Length;
                     }
 
                     if (nextFrame.MpegVersion != firstFrame.MpegVersion ||
                         nextFrame.MpegLayer != firstFrame.MpegLayer ||
                         nextFrame.SampleRate != firstFrame.SampleRate)
                     {
-                        return false;
+                        return nextFrameOffset + firstFrame.FrameLength >= input.Length;
                     }
 
                     previousFrame = nextFrame;


markheath:
For the record, this also resolves #203, which reported the same "wrong sample rate inferred from non-audio data at the start of the stream" symptom. The combination of `ffda548` (multi-frame confirmation) and this end-of-file fix covers that case too.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1260, opened 2026-05-01
Modernization of the NAudio Library

mdcdi1315:
As I started in discussion #1250, this PR starts as a effort to modernize further the library as I recently saw activity here.

This is nothing finalized, nothing stablized, NOTHING.

As such, feel free to modify anything herein. Anything you feel that is off, please feel free to remove and/or comment alternatives.

On the works:

1. Implementing `IMFByteStream` wrapper instead of using current `ComStream`

2. Provide API around audio tags - possibly this will an `ITag` inteface containing common properties and some tag readers - ID3 version 2, MP4 and FLAC.

3. Provide API and source generator for data structures - will allow faster data structure developing through simply using fields and attributes. We always read complex data through structures throughout the codebase - so let's make our life a bit easier.

4. `WaveFormat` abstraction (This has already begun by some of my commits of this PR, please approve or request to modify to apply further modifications)

5. Implement MacOS wrappers (Investigating at these since 2 years before, had created a working prototype but abandoned it back then due to limited time)

Have some more stuff on my mind, but those are for much later.

Note: Will later make individual PR's on the next mentioned features that I am implementing, just want for now to ensure that I will complete the Core library to a certain extent. I want to add more here for sure, but I will for now get involved with the Core stuff, and at the same time the number 3 from the above list.


mdcdi1315:
Hey @markheath have you seen this? I cannot proceed unless you approve.
Your vision of the library API may be different in what I do see.
Even a simple no or yes does make the difference for me.
I have just marked it as draft because there are stuff that will be added bit by a bit in coordination by you, not once.

markheath:
I've not had time to have a proper look yet, but will try to get round to it. As I mentioned I have been doing some very major rearchitecture of NAudio myself (see on the naudio3dev branch) to see what is possible, and have got a number of plans myself for various enhancements to the APIs that modernize things while minimizing breaking changes. I do need to warn that I rarely accept major rearchitecting PRs for NAudio as I need to be able to maintain this project long-term so I need to clearly understand what's going on and agree with the rationale behind each design and naming decision. The most interesting part to me of what you've done is your approach to IAudioFormat. I have not got to the point of making decisions on what a future AudioFormat that's not tied to WAVEFORMATEX would look like, and the interface that WaveFormat still implements is an interesting one, although would need to think about how that works with various interop scenarios. At the moment I'm chasing down a tricky access violation that is probably in ACM so that's slowed me down quite a bit recently.

It would be good for you to look at the naudio3dev branch. It's not yet been reorganized into a new assembly structure which will likely keep all Windows functionality separate from the cross-platform functionality, and maybe segregate things a bit more cleanly (e.g. not mixing DMO, Wasapi and MF all in the same assembly), but once that's been done I'm hoping to be able to publish some "alpha" NuGet packages to get some early feedback

mdcdi1315:
Thanks for replying blazingly fast! Really, didn't expected it and I appreciate that :-)

> I have not got to the point of making decisions on what a future AudioFormat that's not tied to WAVEFORMATEX would look like, and the interface that WaveFormat still implements is an interesting one, although would need to think about how that works with various interop scenarios. At the moment I'm chasing down a tricky access violation that is probably in ACM so that's slowed me down quite a bit recently.

Just a rough sketch. I am looking in that too. In early I was thinking something like a static class with a dictionary that keeps mappings to a second converter interface that would allow to de/encode from a type pointer. But it is a thing that will definitely need planning and coordination. Another thought was the already good and old Serialize method that you already have in V2, but possibly declaring directly `Stream` on the input parameter.

Side-note about ACM: Let's leave it behind for now. It is best to just continue on the modern API's and later on a Windows Legacy assembly could be created.

For now I want to improve the stream utils by adding some alignment and padding methods, and some more for writing and loading enums. First tests to me however show around 2 ms on same endianness when loading numbers running 7 different memory stream de/encodes, so this gets really good! (And then get migrated out of BinaryReader of course because it does not provide flexibility at all)

For now these are my goals:

-> Cleaning up the utils in general, remove bloat and add useful Span manipulations

-> Finding a way to abstract `WaveFormat`

-> Add the Data Structure API

-> Windows: `IMFByteStream` wrapper for .NET to avoid using `ComStream`

Hope that these points cover you.

mdcdi1315:
Oh, forgot to tell that abstracting `WAVEFORMATEX` away does not mean that is subject of deprecation.

Due to your general decisions, I want to make all these updates to not break the library, but rather become the key components in the future.

Oh, and that's why I am contacting you in first place. To see what you expect and perform changes as needed. I do not even expect this branch to be merged at some time.

mdcdi1315:
Have closed it for now, will continue experimenting outside of it, removing IAudioFormat changes and only modifying the Core library stuff, due to the branch naming changes you have done. And also Git is very confusing for me times to times.
Will open soon a new PR with only the Core stuff and the IMFByteStream wrapper, otherwise only those I wanted to be merged for now.

PR #1288, opened 2026-05-16
Introduce IMFByteStream wrapper, instead of using IStream -> IMFByteStream way

mdcdi1315:
## Summary

This PR aims to remove the IStream wrapper that is causing issues due to the shim itself.

Additionally, this PR adds a `CommonHResults` class to the Core library, needed for translating the .NET `Stream` -> `IMFByteStream` calls. It does provide some common `HRESULT` constants, such as `E_INVALIDARG` or `E_FAIL`.

Originally defined in PR #1260 .

## Testing

Using `NAudioConsoleTest`.
But it is really flaky and needs modification. 
This happens because the source reader cannot discover, for some reason, the file type automatically, and it has to be specified through `IMFAttributes` on the wrapper. Don't know how the shim specifies that however (One guess could be that the shim itself discovers it before providing bytes to the reader).
However, I do not know how the native IMFAttributes could be disposed of in the wrapper. I am not so sure whether both the Sink Writer and the Source Reader call the `IMFByteStream.Close` method reliably.
For the above, I might implement custom and managed IMFAttributes class.


markheath:
Thanks for this contribution. I will need time to look into this in more detail, but I did use AI to help me understand the big picture and have generated this feedback for you...


Removing the `Stream → IStream → MFCreateMFByteStreamOnStream → IMFByteStream` chain in favour of a direct `IMFByteStream` wrapper is the right direction. Owning the implementation end-to-end gets rid of the documented IUnknown-vs-IStream QI hazard and our reliance on the opaque Microsoft shim, and `CommonHResults` is a useful addition in its own right.

A few things I think need to be resolved before this can land:

**1. Content-type discovery (the flakiness root cause).** The Source Reader can't auto-detect the container because the new wrapper supplies no content-type hint, whereas the removed MS shim evidently did. This is the core unfinished piece: the wrapper needs to expose `IMFAttributes` carrying `MF_BYTESTREAM_CONTENT_TYPE` / `MF_BYTESTREAM_ORIGIN_NAME` (or sniff the leading bytes). Until that's in, the read path is unreliable rather than just "flaky."

**2. Stream ownership/lifetime change.** `ComStream` disposes the underlying stream on `Dispose()`/`Close()`; `MFByteStreamFromStream` explicitly does not. Switching `StreamMediaFoundationReader` and `MediaFoundationEncoder.Encode(Stream...)` over silently changes who closes the caller's stream. We need to either preserve the old ownership semantics or audit/adjust every caller and call it out in release notes — right now it's an undocumented behaviour change.

**3. Possible native buffer overrun in `Read`.** `ArrayPool<byte>.Shared.Rent(cb)` can return an array larger than `cb`. `stream.Read(rented)` then fills up to `rented.Length`, and `read` (which can exceed `cb`) bytes get copied into the native `pb` buffer that's only `cb` bytes — a native overrun whenever the stream returns more than `cb`. The read needs to be bounded to `cb`.

**4. Async `EndWrite` looks backwards.** It copies the (uninitialised, pooled) managed buffer *into* the native pointer and then writes that managed buffer to the stream. For a write it should copy native → managed, then `stream.Write`. The synchronous `Write` does this correctly; the async path appears to write garbage. Worth adding a round-trip test that specifically exercises the async `Begin/EndWrite` path.

**5. `CommonHResults.E_STRING_NOT_NULL_TERMINATED`.** It's typed `uint` (every other constant is `int`) and the literal `0x80000017E` is 9 hex digits — it overflows 32 bits. The intended HRESULT is `0x80000017`.

**6. `Close()` reliability.** The "reset `Position` to the init value" trick is noted as flaky, and it depends on both the Source Reader and Sink Writer reliably calling `Close()`. Worth confirming that assumption (and how the native `IMFAttributes`, once added, gets released) before relying on it.

Happy to help review the follow-ups — particularly the attributes/content-type piece, since that's what unblocks reliable reads. Note that `MediaFoundationReaderTests.CanRoundTripStreamThroughMediaFoundationCcwPath` covers exactly this path but is `[Category("IntegrationTest")]` and excluded from the headless CI run, so this code is effectively untested in CI today — might be worth a headless-safe round-trip test as part of this work.

mdcdi1315:
First of all, thanks for the response.
Okay, sure about Number 3 in the list.
I will get later to the buffer overruns and will add some test to all the methods to read and write some bytes for the reliability.
About Number 5, it must be a left-over from my own code, will change it to `unchecked((int))`. This happens because in my own code, I have defined `HRESULT` to be `System.UInt32`, which it is correct in the native code (Headers typedef the `HRESULT` to be a `DWORD`, which in turn the `DWORD` corresponds to `System.UInt32`). Should I also deprecate those constants being on the `HResult` class too and hardwire them to the new `CommonHResults` class?
About Number 2, I frankly do not know what you want to do with that. Leaving it to your own discretion.
About Number 1, I still look on how to implement it. Am currently finding all the possible audio file headers I can, so that to test them all and generate the required MIME type. But sure, always some assistance helps the process.

mdcdi1315:
Fixed 2,3,4,5 and 6 from the list.
About 2, I decided to call dispose explicitly from the `MediaFoundationReaderFromStream` class. If so needed for the encoder as well, let me know so that to create a commit for it.
About 6, I decided to not trust `Close` method at all, see my comments on that method.

Apart from that, I changed a bit the interop code and fixed the EndRead/Write methods to also report status now.

markheath:
Thanks, this seems like this is moving in the right direction. Several of the issues are resolved but a few more things worth looking at. I again made use of AI to help me review where we're at, and weighed up a few options for HResults, and got it to generate a report afterwards (see below). Happy to discuss any issues further


**Confirmed resolved:**

- **Read buffer overrun (#3):** the move to `new Span<byte>(pb.ToPointer(), cb)` / `ReadOnlySpan<byte>` is exactly right — bounded to `cb`, zero-copy, and simpler than the pooled-buffer version. 
- **`EndWrite` direction (#4):** now writes from the native pointer into the stream, and both `EndRead`/`EndWrite` report status via `pResult`. Good.
- **`Close()` (#6):** I like this approach a lot more than the original — making `Close()` a no-op with the explanatory comment, moving the reposition into an explicit `ResetPosition()` that the reader calls deterministically, and releasing the native attributes through an explicit `Dispose()`. That's a clean ownership model.
- **Stream ownership (#2):** disposing the wrapper + stream from `StreamMediaFoundationReader.Dispose` preserves the historical behaviour, and leaving the encoder's output stream alone matches what the old `ComStream` path did there too, so the asymmetry is actually correct — no change needed, I just want to make it a conscious, documented decision rather than incidental.

**Still outstanding on design/implementation:**

1. **Content-type discovery — this is the gating item.** Exposing `IMFAttributes` on the wrapper and adding `MfByteStreamAttributes` is the right plumbing, but nothing ever populates `MF_BYTESTREAM_CONTENT_TYPE` — the store is created empty, so the source reader still has no hint and the original flakiness remains unaddressed in practice. The header-sniffing → set-content-type piece you mentioned you're working on is what actually closes this out. Happy to help brainstorm the format-detection table if useful.

2. **`CommonHResults` placement.** To answer your question directly: I'd rather these live on the existing public `HResult` class than introduce a second public generic HRESULT type in the same `NAudio.Utils` namespace. Right now we'd ship two public types that both define `S_OK`/`S_FALSE`/`E_INVALIDARG`, and the `E_INVALIDARG` values even diverge (`HResult` has the legacy `0x80000003`, your set has the canonical `0x80070057`) — that's a footgun depending on which `using` resolves. Proposal: fold the new constants into `HResult`, take the opportunity to correct `HResult.E_INVALIDARG` to `0x80070057`, and fix the `E_STRING_NOT_NULL_TERMINATED` literal (it's still 9 hex digits — `0x80000017E` truncates to `0x17E`; should be `0x80000017`). Keep the domain-specific sets (`DmoHResults`, `AudioClientErrorCode`, `MediaFoundationErrors`) where they are — those are correctly scoped. 

3. **Async error propagation.** `EndRead`/`EndWrite` set the status on `pResult` but always `return S_OK`. A caller checking the `EndRead` return value will see success with `pcbRead = 0` on a failure, i.e. treat it as a clean EOF rather than an error. Worth making the return value reflect the operation result.

4. **Async slot leak on failure.** If `PutWorkItem` fails after the `Interlocked.CompareExchange` has set `asyncdata`, the slot is never cleared and every subsequent `BeginRead`/`BeginWrite` returns `E_ILLEGAL_METHOD_CALL` permanently. Resetting `asyncdata` on the `PutWorkItem` failure path would fix that.

5. Minor: in `Startup()`, if `CreateWorkQueue` throws after `MFStartup` has already succeeded, `initialized` stays `false` and a later `Startup()` will start MF a second time. Worth ordering so MF isn't left half-initialised.

mdcdi1315:
Thanks for the response again, to conclude, here is what I am going to do in the next days:

1. Remove `CommonHResults`, get all constants to `HResult`, and fix `E_INVALIDARG`. About the `E_STRING_NOT_NULL_TERMINATED` constant, I want to verify this in the native Windows header. But there is possibly a sneaky typo.

2. About the Content Type progress
>  Content-type discovery — this is the gating item. Exposing IMFAttributes on the wrapper and adding MfByteStreamAttributes is the right plumbing, but nothing ever populates MF_BYTESTREAM_CONTENT_TYPE.

Now that I was testing with the latest commit, I cannot make the source reader recognize any file - it fails to all of them anymore. And even if I specify content type and origin name, it fails even with those hints. For example, explicitly specified  `audio/mp3` and a random file URL pointing to a fantastic .mp3 file, specify a MP3 file and still not recognizing it. (However, it somehow this works on my own wrapper code, which is driving me crazy!)

I did not intentionally put yet any code referencing content type because I need to resolve the above first - always want to test something and know that it works before proceeding on.

But I will first check on your other suggesstions, maybe it is such a sneaky problem with the EOF that I did not knew (Microsoft docs do not mention explicitly the EOF part), but I am almost 99% certain that that is not that. And since the docs say nothing about it I assume that the caller knows that whence 0, end of stream. Note also that EOF can be also discovered by IsEndOfStream method. Also, error handling on async path happens on the result object, not on EndRead/Write, if I am correct.


3. All other minor issues will be fixed as well, after I manage to overcome the concerns added on Number 2.

4. About the file type inspection, I think that it could make it to the Core library. That is, having for example an AudioFileTypeInspector class in the `Utils` namespace, and having a single static method `Resolve` that accepts a `Stream` and gets a result which file type it is. Share with me your considerations about this specific idea, will push later on a commit for just a review and first opinion.

mdcdi1315:
I think that I am almost ready.

Only task left now is to delete the old ComStream class and everywhere it is still referencing and add more file format finders.

If you find anything else you see wrong, feel free to comment.

Otherwise, I make and push the last commit of removing the `ComStream` and related stuff.

markheath:
Great, I think this is moving in the right direction and I like the HResult changes. Again, I've reviewed this with AI and added in a few points of my own.

**Resolved nicely:**

- **HResult consolidation:** exactly what I was hoping for — `CommonHResults` folded into the existing public `HResult`, `E_INVALIDARG` corrected to `0x80070057`, `E_STRING_NOT_NULL_TERMINATED` to `0x80000017`, old `MAKE_HRESULT` kept as an `[Obsolete]` forwarder, and the new `IsError` helper is a nice touch. No further changes needed here.
- **`Startup()` ordering:** the `MFStartup` → work-queue → `catch { MFShutdown() }` sequencing (and the symmetric `Shutdown()`) is correct now.
- **Async error propagation:** `EndRead`/`EndWrite` returning `GetStatus()` rather than a hard-coded `S_OK` is the right behaviour.
- **Content-type:** the finder is now actually wired into the constructor and sets `MF_BYTESTREAM_CONTENT_TYPE` — that's the gating piece finally in place.

A few things still need resolving before this can land — roughly in priority order:

**1. Async object lifetime looks unsafe — I suspect this is what's behind the intermittent failures you're seeing.** In `MFByteStreamFromStream.CreateAsyncCall`, the `finally` releases both async-result objects (`ComActivation.ReleaseBoth(result2, ptr2)` / `ReleaseBoth(result, …)`) immediately after `MFPutWorkItemEx` returns — i.e. before the work queue has run the item that still references them. That's a use-after-free, and it lines up well with the "works in my own app, fails here, ~30% on FLAC" symptom. Relatedly: `EndRead` looks the size up by the `IMFAsyncResult` pointer MF passes it, while `SetAsyncCallSize` stores it under the pointer from `result.GetObject()` — COM doesn't guarantee those `IntPtr`s share identity, so that lookup can silently miss and give zero-byte reads. And the `Monitor.Enter(this)` on the COM-exposed instance, combined with the now-`MultiThreaded` work queue, is a deadlock risk. Given MF uses the synchronous `Read` path for local seekable streams in the common case, one option worth considering is returning `E_NOTIMPL` from the `Begin*/End*` methods until there's a concrete need and a verified design — it's a legal `IMFByteStream` choice and would remove a large, fragile surface. Happy to talk this through.

**2. The MIME strings are wrong, so #1 probably isn't actually fixed yet.** `MimeType => $"audio/{MimeTypeName}"` produces `audio/mp3` / `audio/mp4` / `audio/flac`. MF's registered byte-stream handler types are `audio/mpeg` for MP3 and `audio/mp4` (or `video/mp4`) for MP4 — `audio/mp3` isn't standard and MF will likely ignore it. This is consistent with your note that it still fails even with an explicit content type. Mapping detected-format → the exact MF content-type string is really an MF-side concern, so I'd keep that mapping out of the Core type and use the correct MF MIME values.

**3. Consider scoping the format-discovery API down for this PR.** Stream sniffing is a reasonable general capability and Core is a defensible home, but `AudioFileFormat` / `AudioFileFormatFinder` as a new **public** extensibility surface — abstract base, builder, singletons, plus a process-wide mutable `static` registry (`AddDefaultFileFormat` over a non-thread-safe `List`, which will throw if mutated concurrently with `FindFormat`) — is a lot of new public API arriving incidentally inside a byte-stream fix. I'd prefer we make this `internal` for now (enough to unblock the content-type hint) and design/ship the public audio-format-discovery API as its own follow-up PR where it can get proper review. That keeps this PR focused and avoids committing to a half-formed public contract.

**4. Naming pass.** Two things: the new code is largely `snake_case` for fields/locals/params (`wrapper_init_pos`, `r_w_calls_async_results`, `read_mode`, `queue_token`, …) whereas NAudio uses camelCase throughout — worth a pass to match the surrounding style. And the acronym casing is inconsistent even within the PR: `MP3FileFormat`/`FLACFileFormat` vs `MfByteStreamAttributes`/`MediaFoundationApi`. Per .NET guidelines these should be `Mp3FileFormat` / `Mp4FileFormat` / `FlacFileFormat`.

The priorities are really #1 and #2; #3 and #4 are about keeping the change reviewable. Well done for the persistence - it might take a few more rounds to get this right, but will be a good addition to the library once it's done.

mdcdi1315:
Thanks again for the response, really.

As far as I am concerned about 1 and 2, the async handling was the issue for number 1, that's why I rewrote it from the beginning. I am aware about the COM's identity, but given the fact that `CreateAsyncResult` creating an `MFASYNCRESULT` structure, it is correct because that one reads directly the field (which it won't ever be changed during it's lifetime).
However, yes I do not also agree with that design, it is very agressive. My other way was porting `MFASYNCRESULT` as well, but for me that is fragile because who does guarantee us that this will exist tomorrow? All in all, what I wanted here is to make it work. I also have looked up the non-async option, but there is not a closely relevant attribute to use on the source reader to disable (Could it be the [`MF_LOW_LATENCY`](https://learn.microsoft.com/en-us/windows/win32/medfound/mf-low-latency) one, but using async or not is clearly a MFT feature, and not something that can be enforced by the source reader itself), because if now someone makes the async methods `E_NOTIMPL`, the reader fails instantly. Note, my comments in code refer to the old implementation that is defined in [`ceb3173`](https://github.com/naudio/NAudio/pull/1288/commits/ceb3173e47c44de89a3a91b4f025d7b0fac22185) commit. They do not apply anymore. Will change them to reflect the changes.

About now number 3: 

This was a rough sketch for the byte stream purpose, and yes will need a review in the close future. Will open in the next days an issue in which will debate the design of the format-discovery API. But I am certain that enforcing builder pattern + singletons is the right way to move with this. For now, I will move everything into the `Wasapi` library to make it internal. And I really want to have this feature in V3 because this is going to be used for another 7 years or more - so we need something stable and something that applications can work with the byte stream content type discovery if Microsoft adds more media sources in the future that the lib has not yet added.

Finally, about number 4, do not worry about that I will convert everything to camel case as you want in the final polishing commit (That will remove `ComStream` and `IStream`). Only I have to remind me to do it, actually :-) .

Will get done by tomorrow Number 3.
About Number 1, I was already thinking something with an index instead to the dictionary, to avoid depending on the result object value for that.
About the lifetime, it is not use after free, since MF should have kept it's data reference to the async result so that the queue knows. Made numerous tests around it, found that only that the first async object must not be released because `CreateAsyncResult` does not `AddRef` them internally. About the free on the first one async result, as you noticed, I intend only to delete the created RCW. About threading, `MultiThreaded` was for the tests - my mistake!

Note also, everything is working, i.e. give a .mp3, .mp4 or .flac to the source reader and it will decompose it to PCM data for you. About the MIME types, Microsoft mentions that those are also correct to use. I have used them now for like a thousand of times or so - and all the files are recognized correctly. However, there is also the [`MFGetSupportedMimeTypes`](https://learn.microsoft.com/en-us/windows/win32/api/mfidl/nf-mfidl-mfgetsupportedmimetypes) function, giving the MIME types supported by the source resolver.
I recommend here to also port this and add to the reader settings a MIME type string to be selected by the users.

To say about the MIME type, here are all the installed MIME types in my own machine:
Screenshot 2026-05-19 215715

> Well done for the persistence - it might take a few more rounds to get this right, but will be a good addition to the library once it's done.

Thanks, I totally had 3 burnouts in these days in tests in order to pinpoint issues and how the threading model should look like.

mdcdi1315:
Note: I have detected mem leaks once I used the Media Foundation reader with the latest, it leaks large sizes (i.e. 65535 bytes).
Just uploaded this to document it so as to not forget it, will test to see what happens with the URL reader and compare results. If the results happen from my wrapper, it is an issue with that one, otherwise it is also an issue with the Source Reader. My first findings indicate and code paths to `RTWorkQ` DLL as well, meaning that my code leaks something as well.

mdcdi1315:
Here is what is done now:

- Fixed unsafe async model used on the byte stream wrapper - now it can be multi-threaded, so I am leaving it that way.
- Fixed some mem leaks on the wrapper, also discovered a leak on `MediaFoundationReader`
   - Opened an issue to take some time in the future to investigate more COM leaks
- Cleaned up my added code on `MediaFoundationApi`.
- Started renaming everything to camel case where needed.

Remaining stuff that are still needed to be done:

- Complete camel case renames
- If possible, fix and the last mem leak on the wrapper (Some 8.192 byte buffers are still leaking, probably from async handling, average 8-10 leaks on a single file read pass)
- Remove `ComStream` and whatever references it (That is for the end)

mdcdi1315:
I believe that the PR is ready for review.

As always, feel free to comment if there is something that blocks the PR.

mdcdi1315:
Hey @markheath this is ready. Have you seen this at all?

Just a yes or no is enough for me.

I am asking because I haven't saw any activity on this and you really wanted this PR to be merged.

Am also saying it because I want to move on modernizing the PR #741 so that it can be available for V3.

markheath:
I'll try to find some time to look at this soon - have got quite a lot of things going on and want to make sure I fully understand this.

Perhaps you can tell me what you've managed to test this with - what types of audio have you played and encoded to with streams, and what sort of streams (file-backed/memory-backed/network streams?)

I'll add some more comments on #741  as I was looking at that earlier this week

markheath:
Just thinking about ownership of the streams. 

If I understand correctly `StreamMediaFoundationReader` owns the input stream and will dispose it for you when you dispose `StreamMediaFoundationReader`. That might match existing behaviour, but not completely sure it's ideal. We don't do it in `WaveFileReader`, `Mp3FileReader` etc - i.e. if we didn't open it, we don't close it. In the .NET framework they have `leaveOpen` overloads for things like `StreamReader` as a different way to address this. So I'm thinking we either don't close the stream passed in to match other NAudio classes or we copy the .NET approach with an optional `bool leaveOpen=false` constructor param.

The encode path also doesn't close the stream. And I think that's correct - `MediaFoundationEncoder.Encode(Stream)` should not close your stream for you - so nothing to change there.

mdcdi1315:
> If I understand correctly StreamMediaFoundationReader owns the input stream and will dispose it for you when you dispose StreamMediaFoundationReader. That might match existing behaviour, but not completely sure it's ideal. We don't do it in WaveFileReader, Mp3FileReader etc - i.e. if we didn't open it, we don't close it. In the .NET framework they have leaveOpen overloads for things like StreamReader as a different way to address this. So I'm thinking we either don't close the stream passed in to match other NAudio classes or we copy the .NET approach with an optional bool leaveOpen=false constructor param.

That is very trivial to be done - there are two options:

1. I push at Monday night a commit that will add a second constructor with a leaveOpen boolean.

2. Leave it as-is, accept the PR and if there is a decision in the future for leaveOpen, it will be a trivial PR on it's own if so decisions are made.

Please feel free to comment me about your final decision on this, once you have made one.

About #741: I am just currently looking at it's code. Have not managed yet to have a proper look at it, but I believe that it needs anyway a re-implementation for V3, and want some things to be improved which in my opinion require improvements, such as how the topology is constructed. 

Once I find additional time, I will open a new PR that will comply with V3 patterns and requirements.

markheath:
OK, make sure you read the comments I added over on #741 as well. I think this PR is pretty close, just need a little more time to finish reviewing it. There are a couple of small things I'd still want to tweak before getting it in, but I could easily just do those myself as a final pass rather than going round the loop lots of times with you on this.

Do let me know what you've tested with this though as that's the other main thing I want to be sure we've covered this sufficiently with a range of use cases (just running stuff through the existing test harnesses is fine - we don't need to write a load more test code).

mdcdi1315:
> Do let me know what you've tested with this though as that's the other main thing I want to be sure we've covered this sufficiently with a range of use cases (just running stuff through the existing test harnesses is fine - we don't need to write a load more test code).

Have been tested a couple of real-world scenarios (i.e. WASAPI playback) with C# interactive and some others as well, but I want to re-run them at some time just to be sure that this new wrapper is now fully safe to replace the current implementation.

> OK, make sure you read the comments I added over on #741 as well.

I will do some research tomorrow why we should also port the Media Foundation player as well. If I do not find anything useful, I will bring the macOS wrappers to the foreground, but that would take much more time as that will need planning and coordination. I also do research again for that, and will make sure to open a FR issue for that, because I want you over to decide what it will be needed and what will not. For what you need to know for now about that, they provide similar API's to Windows with almost the same strengths, and I have a test device for this so no worries for now.

NAudio.Wasapi/MediaFoundation/MediaFoundationHelpers.cs
```diff
@@ -23,7 +25,9 @@ namespace NAudio.MediaFoundation
     /// </remarks>
     public static class MediaFoundationApi
     {
-        private static bool initialized;
+        // Let's be sure that the static initializer assigns the below variables to correct initial values.
+        private static bool initialized = false;
+        private static uint queue_token = 0U;
```
markheath: Use camelCase not snake_case

NAudio.Wasapi/MediaFoundation/MediaFoundationHelpers.cs
```diff
@@ -33,11 +37,53 @@ public static void Startup()
...
+        }
+
+        /// <summary>
+        /// Shuts down Media Foundation
+        /// </summary>
+        public static void Shutdown()
+        {
+            if (initialized)
+            {
+                try
```
markheath: I think it might be better to release the work queue before calling MFShutdown as you're not supposed to call any mfplat functions after MFShutdown (and it might throw atm)

NAudio.Wasapi/MediaFoundation/MediaFoundationHelpers.cs
```diff
@@ -33,11 +37,53 @@ public static void Startup()
             if (!initialized)
             {
                 MediaFoundationException.ThrowIfFailed(
-                    MediaFoundationInterop.MFStartup(MediaFoundationInterop.MF_VERSION, 0));
-                initialized = true;
+                        MediaFoundationInterop.MFStartup(MediaFoundationInterop.MF_VERSION, 0)
+                );
+                try
+                {
+                    queue_token = CreateWorkQueue(WorkQueueType.MultiThreaded);
```
markheath: this is always creating a work queue every time we use MF, even if we never need to use one.
There are a couple of options here - we could allocate it lazily  (e.g. on the first BeginRead/BeginWrite)
But I believe that MF offers standard shared queues you can MFPutWorkItem onto without allocating your own one - e.g. take a look at MFASYNC_CALLBACK_QUEUE_MULTITHREADED. That might be simpler and removes the need to allocate and release your own work queue, as well as the need to track the queue token

NAudio.Wasapi/Utils/FileFormatDiscovery/AudioFileFormat.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
+using System.IO;
+using System.Runtime.CompilerServices;
+
+#pragma warning disable CS0659 // AudioFileFormat is not overriding GetHashCode
+
+namespace NAudio.Utils.FileFormatDiscovery
```
markheath: put these in the MediaFoundation rather than Utils namespace for now. If we promote this to a first-class concept later we'll put them back in utils.

NAudio.Wasapi/MediaFoundation/IMfByteStreamAsyncCallSizeHandler.cs
```diff
@@ -0,0 +1,31 @@
...
+    [GeneratedComInterface]
+    [Guid("af1b8de9-d41e-4e22-b17b-5a3cd4ed75f7")]
+    [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
+    internal partial interface IMfByteStreamAsyncCallSizeHandler
+    {
+        [PreserveSig]
+        int GetDataSize();
+
+        [PreserveSig]
+        void SetDataSize(int data_size);
```
markheath: dataSize

markheath (commented):
Pretty close now- just a couple of minor cleaning up naming/namespace issues. I had to do a bit of learning about workqueues to help me understand this and it seems to me that we can perhaps simplify slightly here and not need to allocate and release a workqueue ourselves but use one of the built-in ones

mdcdi1315:
> Pretty close now- just a couple of minor cleaning up naming/namespace issues. I had to do a bit of learning about workqueues to help me understand this and it seems to me that we can perhaps simplify slightly here and not need to allocate and release a workqueue ourselves but use one of the built-in ones

Will do the last renames, possibly missups while cleaning up, and will make sure to put the stream sniff API to `MediaFoundation` instead.

About the work queue I was just thinking of allocating one of our own since the default multithreaded queue could be used in multiple processes but I recently found that is per-process, so I will possibly fix that. Even Microsoft suggests using the default work queue.

mdcdi1315:
Ahh, should I also port MMCSS API for improved scheduling of MF work queues?

Am specifically talking for [`MFRegisterPlatformWithMMCSS`](https://learn.microsoft.com/en-us/windows/win32/api/mfapi/nf-mfapi-mfregisterplatformwithmmcss) and [`MFUnregisterPlatformFromMMCSS`](https://learn.microsoft.com/en-us/windows/win32/api/mfapi/nf-mfapi-mfunregisterplatformfrommmcss) functions.

BTW, make sure to see at some time the comments I added at PR #741.

markheath:
No, I don't think we need MMCSS for this work - just keep it simple and let's get this PR completed. I think the built-in multithreaded queue with default scheduling is fine for this work.

mdcdi1315:
> No, I don't think we need MMCSS for this work - just keep it simple and let's get this PR completed. I think the built-in multithreaded queue with default scheduling is fine for this work.

Okay, I will push the commit in a bit. I was fixing some naming stuff anyway, so I am close to complete it.

mdcdi1315:
Made now the final tests around the reader, seems that it is OK and a simple WASAPI playback was normally completed.

Am also suggesting to add an 'enhancement' label to this PR as it is such.

markheath:
I just realized I never got back to you about stream ownership for `StreamMediaFoundationReader`. I think we do need to keep it consistent with the other NAudio WaveStreams that don't close streams they didn't open.

So the two options are:
(1) - simply don't dispose the stream you were passed in `StreamMediaFoundationReader.Dispose`
(2) - add an optional `leaveOpen` parameter with a default value of `true` 

Or if you prefer we can complete this and I will make the change afterwards in a separate PR.

Let me know what you want to do. I will also clone this PR locally and test a few files through it myself as a final check.

mdcdi1315:
I think that implementing option 2 now is a bit inconsistent - because all other readers currently do not support this and the `StreamMediaFoundationReader` would be the first one to do that. You had mentioned it before this, and I thought about it.

Best way to move here is the following: Since the current impl is backwards compat, we keep it as-is. I am saying that because if it is to make any change now and add the parameter in the ctor we would have to open a PR for all the other reader ctors as well and apply the change there too.

So, if it is to add the ctor anyway, we should add it in a upcoming PR in which we will adopt this to all the readers.
So, if we finally decide to add the parameter I can handle that PR.

markheath:
I've done some testing on my machine and all seems to work fine - we're very close now.

Not sure what you mean by saying the existing behaviour is backwards compat? The original `StreamMediaFoundationReader` didn't close the stream for you and this implementation does. And other NAudio stream readers (WaveFileReader, Mp3FileReader, AudioFileReader) also don't close the stream if they didn't open it. 

So to maintain original behaviour and consistency with the rest of the library I still think we want option 1 - no dispose of the stream you were passed.

Apart from that, this is ready to merge.

Also just quickly checking - did you solve the memory leak issue you mentioned earlier?

mdcdi1315:
> Not sure what you mean by saying the existing behaviour is backwards compat?

Oh, I thought it was backwards compat due to `ComStream` wrapper. Anyway, I will remove it.

> Also just quickly checking - did you solve the memory leak issue you mentioned earlier?

It probably was the [`sizeHandlerCcw`](https://github.com/mdcdi1315/NAudio/blob/main/NAudio.Wasapi/MediaFoundation/MFByteStreamFromStream.cs#L336) on the `CreateAsyncCall` method, which is the one you fixed yourself.

I have tested it numerous times now, it does not seem to have any other mem leaks and have verified myself all the possible code paths I can think of a couple of times. Even the most recent tests I made did not indicate any new mem leaks.

markheath:
We're done - thanks for your hard work and willingness to iterate on this

mdcdi1315:
My pleasure to help!

I am possibly moving now to my next PR about macOS wrappers.
Will open an issue in the next days about it in which I will describe in more detail what is happening with that, and what are the specifics about it.

For now, I am still sweeping the repo for finding mem leaks and any other missing stuff.

markheath:
Yes, that would be a good high-value addition to the library. Let's open an issue about it and discuss as I have a few thoughts which I can share so we can be aligned on scope and direction

PR #1323, opened 2026-05-29
V3: Fix `IMFByteStream` wrapper invalidly calling `FindFormat` on write mode

mdcdi1315:
## Summary

This PR fixes the byte stream wrapper that was calling to find the format of a stream that was intended for writing.

The fix is to add a boolean parameter to the wrapper that controls whether to call the find format method and by default it has the `true` value which indicates that the wrapper is to be used by readers.

The current code could cause the encoder to fail, if someone was passing a write-only stream to `EncodeXXXX` methods.

## Release notes

This does not possibly need a release notes entry since the byte stream wrapper is a new feature for V3.
But if you feel that this does need an entry, feel free to add one later.

## Testing

This was actually something slipped through the tests - passing a writeable stream only could cause the wrapper to fail.

It slipped because the tests are using `MemoryStream`, which it's `CanRead` property has the value `true`.

### DIFF
--- a/NAudio.Wasapi/MediaFoundation/MFByteStreamFromStream.cs
+++ b/NAudio.Wasapi/MediaFoundation/MFByteStreamFromStream.cs
@@ -26,18 +26,21 @@ namespace NAudio.MediaFoundation
         private IntPtr nativeAttributesPtr;
         private IMFAttributes nativeAttributesRcw;
 
-        public unsafe MfByteStreamFromStream(Stream stream)
+        public unsafe MfByteStreamFromStream(Stream stream, bool readMode = true)
         {
             this.stream = stream;
             try { wrapperInitialPosition = stream.Position; } catch { wrapperInitialPosition = 0; }
             (nativeAttributesPtr, nativeAttributesRcw) = MediaFoundationApi.CreateAttributes(1);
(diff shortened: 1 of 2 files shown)


NAudio.Wasapi/MediaFoundation/MFByteStreamFromStream.cs
```diff
@@ -26,18 +26,21 @@ internal sealed partial class MfByteStreamFromStream : IMFByteStream, IMFAttribu
         private IntPtr nativeAttributesPtr;
         private IMFAttributes nativeAttributesRcw;
 
-        public unsafe MfByteStreamFromStream(Stream stream)
+        public unsafe MfByteStreamFromStream(Stream stream, bool readMode = true)
```
markheath: looks good, although one small suggestion is would it be better to call this `detectContentType` to express its purpose? maybe in the future you might want a readable MfByteStreamFromStream that doesn't need to do context sniffing (e.g. you already know the mime type or you have a forward-only stream)

mdcdi1315: Yes, the name `detectContentType` is better. Will fix it.

PR #1397, opened 2026-07-29
Eliminate redundant input copies in ASIO capture path

markheath:
## Summary

Removes wasted per-channel input copies in the `AsioDevice` duplex and recording capture paths, addressing the concern raised in [discussion #1389](https://github.com/naudio/NAudio/discussions/1389#discussioncomment-17826824).

Previously, every selected input channel was converted native→float **eagerly, once per buffer-switch**, into a library-owned staging buffer — regardless of whether the processor ever read it. When the driver's native input format is already `Float32LSB` that "conversion" is a pure `memcpy`, and when a channel is read only via the zero-copy `RawInput` escape hatch (or not read at all) the whole conversion is dead work. Both scale with channel count, which matters for the high-channel-count scenarios in the discussion.

Two stacked commits:

1. **Float aliasing (low-risk).** When `InputFormat == Float32LSB`, `AsioProcessBuffers.GetInput` / `AsioAudioCapturedEventArgs.GetChannel` now return a `ReadOnlySpan<float>` aliasing the driver buffer directly instead of a staged copy. No new lifetime or safety boundary: the span was already callback-scoped, and input buffers are driver-filled before the callback and only read during it, so an aliased read yields the same bytes a start-of-callback snapshot would. The pointer→float reinterpret is the same one `ConvertFloat32Lsb` already performed.

2. **Lazy conversion (generalizes to all formats).** For non-float formats the native→float conversion is deferred to first `GetInput`/`GetChannel` access and cached for the rest of the callback via a per-channel `InputConverted` guard. Channels touched only via `RawInput`, or never read, are never converted — mirroring how the output side uses `OutputRawAccessed` to skip work. The guard array is allocated once at `Init` and only element-reset each buffer-switch (no per-callback GC), and the reset at the top of every buffer-switch is what prevents serving stale audio from a prior callback.

Behaviour and the callback-scoped span lifetime are unchanged for callers. Non-float formats produce identical converted values; `unsafe` on `GetInput`/`GetChannel` is an implementation detail that does not propagate to consumers (the signatures return `ReadOnlySpan<float>`, not pointers — consumers need no `unsafe` context, exactly as with the existing `RawInput`/`RawOutput`).

## Release notes

## Testing

- Built `NAudio.Asio` and `NAudio.Windows.Tests` with `-p:EnableWindowsTargeting=true` — both clean, 0 warnings.
- Added unit coverage in `AsioFloatInputAliasingTests` (hand-built `AsioCallbackContext`, no hardware, following the existing `AsioOutputOnlyDuplexTests` pattern):
  - `GetInput`/`GetChannel` for `Float32LSB` read the driver's native samples and leave the sentinel-filled staging buffer untouched (proving the copy is skipped), and reflect a post-fetch mutation of native memory (proving a live alias, not a snapshot).
  - The non-float lazy path is driven across two simulated buffer-switches plus a within-callback re-read, asserting convert-on-first-access, cache-within-a-callback, and refresh-after-reset (the stale-audio risk).
  - A regression guard that the non-float path invokes the converter rather than aliasing.
- The NUnit tests run on Windows CI (Windows TFM); as a cross-check I also ran a standalone simulation of the exact guard/reset flow against the real conversion formula, and all three properties held.

https://claude.ai/code/session_015XgkRvizPkWUGyqdiTuwHu

---
_Generated by [Claude Code](https://claude.ai/code/session_015XgkRvizPkWUGyqdiTuwHu)_


markheath:
Have tested ASIO recording through NAudio demo, still working

PR #1421, opened 2026-08-27
Use the Lock class

Seb-stian:
## Summary

This PR replaces private and internal lock `object`s with the [Lock class](https://learn.microsoft.com/en-us/dotnet/api/system.threading.lock?view=net-10.0) which offers [better performance](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/lock#guidelines).

## Release notes

## Testing

Unit tests pass the same as on the master branch.

### DIFF
--- a/samples/NAudioWpfDemo/WasapiRecorderDemo/WasapiRecorderViewModel.cs
+++ b/samples/NAudioWpfDemo/WasapiRecorderDemo/WasapiRecorderViewModel.cs
@@ -1,9 +1,10 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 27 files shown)


markheath:
Thanks, seems like this will make a small perf improvement. We should probably include AcmInterop.AcmLock as that is the main one that this PR doesn't cover.

(note - don't touch DirectSoundOut.m_LockObject - that uses Monitor.TryEnter so can't use Lock).

Seb-stian:
> We should probably include AcmInterop.AcmLock as that is the main one that this PR doesn't cover.
> 
> (note - don't touch DirectSoundOut.m_LockObject - that uses Monitor.TryEnter so can't use Lock).

I will be away from my workstation until the 8th, then I can do the necessary changes. Otherwise feel free to take over!

PR #1422, opened 2026-08-27
Use Regex source generators

Seb-stian:
## Summary

Replace Regex constructors with source generators for [better performance](https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-source-generators).

## Release notes

## Testing

SFZ unit tests are green.


markheath:
thanks - again a small perf win. I think we should also remove the `RegexOptions.Compiled` from the attributes as they are not needed

PR #1423, opened 2026-08-27
Optimize code using Spans

Seb-stian:
## Summary

Use `stackalloc`, the `static ReadOnlySpan<T> =>` pattern and other Span-based APIs for better performance.

~~The `static ReadOnlySpan<T> =>` pattern allocates `System.RuntimeFieldInfoStub` in Debug mode, causing some allocation tests to fail. Those tests now run `Assert.Ignore` when in Debug mode.~~

## Release notes

## Testing

Unit tests pass, with a caveat (see Issues above).

### DIFF
--- a/src/NAudio.Alsa/AlsaOut.cs
+++ b/src/NAudio.Alsa/AlsaOut.cs
@@ -17,12 +17,12 @@ namespace NAudio.Wave.Alsa;
 /// </remarks>
 public sealed class AlsaOut : IWavePlayer, IWaveLatency
 {
-    private static readonly PCMFormat[] PreferredFormats =
-    {
+    private static ReadOnlySpan<PCMFormat> PreferredFormats =>
+    [
         PCMFormat.SND_PCM_FORMAT_FLOAT_LE,
         PCMFormat.SND_PCM_FORMAT_S16_LE,
         PCMFormat.SND_PCM_FORMAT_S24_3LE,
-    };
+    ];
 
     private readonly object sync = new();
     private readonly ManualResetEventSlim resumeGate = new(initialState: true);
--- a/src/NAudio.Core/Codecs/ALawDecoder.cs
+++ b/src/NAudio.Core/Codecs/ALawDecoder.cs
@@ -12,7 +12,7 @@ public class ALawDecoder
     /// <summary>
     /// only 512 bytes required, so just use a lookup
     /// </summary>
-    private static readonly short[] ALawDecompressTable = new short[256]
+    private static ReadOnlySpan<short> ALawDecompressTable => new short[256]
     {
          -5504, -5248, -6016, -5760, -4480, -4224, -4992, -4736,
          -7552, -7296, -8064, -7808, -6528, -6272, -7040, -6784,
--- a/src/NAudio.Core/Codecs/ALawEncoder.cs
+++ b/src/NAudio.Core/Codecs/ALawEncoder.cs
@@ -1,4 +1,6 @@
-﻿namespace NAudio.Codecs;
+﻿using System;
+
+namespace NAudio.Codecs;
 
 /// <summary>
 /// A-law encoder
@@ -7,7 +9,7 @@ public static class ALawEncoder
 {
     private const int cBias = 0x84;
     private const int cClip = 32635;
-    private static readonly byte[] ALawCompressTable = new byte[128]
+    private static ReadOnlySpan<byte> ALawCompressTable => new byte[128]
     {
          1,1,2,2,3,3,3,3,
          4,4,4,4,4,4,4,4,
--- a/src/NAudio.Core/Codecs/MuLawDecoder.cs
+++ b/src/NAudio.Core/Codecs/MuLawDecoder.cs
@@ -12,7 +12,7 @@ public static class MuLawDecoder
     /// <summary>
     /// only 512 bytes required, so just use a lookup
     /// </summary>
-    private static readonly short[] MuLawDecompressTable = new short[256]
+    private static ReadOnlySpan<short> MuLawDecompressTable => new short[256]
     {
          -32124,-31100,-30076,-29052,-28028,-27004,-25980,-24956,
          -23932,-22908,-21884,-20860,-19836,-18812,-17788,-16764,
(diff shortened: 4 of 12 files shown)


markheath:
OK, I did some reviewing of this change earlier with help of AI, and also re-went over it. I don't think we actually need to block the allocation tests in debug mode if we make a couple of minor changes. Anyway - here's a Claude code generated summary of some points to address.

---
### What was verified

The `static ReadOnlySpan<T> => [...]` conversions do what they claim. Checking the IL of every converted getter in the Release build, all of them get the RVA data-section treatment — no `newarr` in any of them, so none allocates an array per access. That includes the `float` tables and `AlsaOut.PreferredFormats`, which is a span of an enum (the case most likely to have quietly degraded). Removing the static constructor and its class-init check from each access is a genuine win, exactly as the asm dump in the description shows.

`BinaryPrimitives` in `XingHeader`, `IndexOfAnyExcept`/`IndexOf` in `ByteArrayExtensions`, and the `sr_table` dead-code removal are all good. `DecodeAsString` preserves the original's subtle `length`-mutating truncate-at-first-null semantics correctly.

Assembly size cost is +1,024 bytes on `NAudio.Core.dll` (+0.3%), which is fine.

---

### 1. `MidiFile` — partial reads (blocking)

`f777bf8` replaces `Encoding.UTF8.GetString(br.ReadBytes(4))` with `br.Read(span)` at `MidiFile.cs:62`, `68`, `100`, `243` and `249`. These are not equivalent:

- `BinaryReader.ReadBytes(4)` loops internally until it has 4 bytes or hits EOF.
- `BinaryReader.Read(Span<byte>)` issues a **single** best-effort read and may legally return fewer bytes.

Against a stream that returns one byte per `Read` call:

```
ReadBytes(4)            -> 4 bytes: 'MThd'
Read(Span[4])           -> 1 bytes: 'M   '
SequenceEqual("MThd"u8) -> False
```

`MidiFile` accepts an arbitrary `Stream`, so a perfectly valid MIDI file loaded from a `NetworkStream`, `DeflateStream`, `SslStream` or any custom stream that returns partial reads will now throw `FormatException("Not a MIDI file - header chunk missing")`. The existing tests use files and `MemoryStream`, both of which satisfy 4-byte reads in one go, so CI stays green.

Suggested fix: `br.BaseStream.ReadExactly(chunkHeader)` (throws `EndOfStreamException` on truncation), or a read loop. Note the return value is currently discarded at lines 62, 68 and 100 (`_ = br.Read(...)`), so a short read isn't even detectable there.

Two smaller points in the same commit:
- `chunkHeader.Clear()` at line 99 is a no-op — the next read overwrites all 4 bytes.
- In the RMID skip loop, `toSkip -= chunk.Length` over-subtracts on the final iteration. It happens to come out right (each read requests `Math.Min(chunk.Length, toSkip)` against the current value, and `toSkip` is `long` so it exits rather than wrapping), but `toSkip -= bytesRead` would be clearer and would also fix the pre-existing assumption that each `Read` returns everything requested.

---

### 2. The two disabled allocation tests

Both failures have a **single** root cause, so disabling two tests overstates the problem. `SamplerEngine` constructs a `ReverbEffect` for its send bus (`SamplerEngine.cs:89`), so the sampler's `Read` path runs the same `ReverbEffect.ProcessBank` loop as the effects test.

That loop uses `CombTuning.Length` and `AllpassTuning.Length` as its bounds. For element types wider than one byte, `static ReadOnlySpan<T> => new T[N]{...}` compiles to `ldtoken <RVA field>; call RuntimeHelpers.CreateSpan<T>`, and the JIT only folds `CreateSpan` to a constant pointer when optimising. In a Debug (minopts) build it stays a real call and each `ldtoken` allocates a `RuntimeFieldInfoStub` of about 36 bytes. That accounts for the numbers exactly: 1024 samples × 2 banks × 14 property reads × ~36 B ≈ 1 MB per `Process`.

Measured on `38f2023` with the `Assert.Ignore` guards removed:

| Test | Debug allocation |
|---|---|
| `EffectAllocationTests.SteadyStateProcessDoesNotAllocate(ReverbEffect)` | 528,482,304 bytes / 512 `Process` calls |
| `SamplerPerformanceRegressionTests.SteadyStateNoteCyclesDoNotAllocate` | 206,438,400 bytes |

Hoisting the two loop bounds out of the spans fixes both:

```csharp
private const int NumCombs = 8;
private const int NumAllpasses = 4;
```

used in `ProcessBank` and `Reset` in place of `CombTuning.Length` / `AllpassTuning.Length`. Applied on top of the current head with both `#if DEBUG` guards removed, the result is `NAudio.Core.Tests` 1498/0 and `NAudio.Sampler.Tests` 205/0 in **Debug**, with the span optimisation retained everywhere including `ReverbEffect`'s own tables.

The argument against the `Assert.Ignore` approach: these two tests are the only automated check of the allocation-free-audio-thread claim, and the documented local dev loop in CLAUDE.md builds Debug, so the guards disable them in exactly the configuration contributors run. `FdnReverbEffect.BaseDelaysMs` is unaffected either way — it's read only from `Configure`, never from the audio loop.

An alternative, if the `const` hoist feels like papering over the pattern: leave `CombTuning`/`AllpassTuning` as `static readonly int[]`. They're 12 ints read once in `OnConfigure`, so there's no measurable win from converting them, and the demonstrated gain in this PR is in the μ-law/A-law decode tables, which are genuinely hot.

---

### 3. `Mp3Frame.LoadFromStream:68` — still allocating

```csharp
Span<byte> headerBytes = new byte[4];
```

This is a heap allocation typed as a span. It's the one place in the PR where a per-call allocation actually matters, because `Mp3FileReaderBase` calls `LoadFromStream` for every frame when building the index. `IsLikelyAudioFrameSequence` at line 124 already uses `stackalloc byte[4]` correctly. The same works here — `headerBytes.CopyTo(frame.RawData)` is fine with a stack buffer.

Related: switching `input.Read(headerBytes, 0, 4)` to `input.Read(headerBytes)` is a small pessimisation for user-supplied `Stream` subclasses that only override the `byte[]` overload, since `Stream.Read(Span<byte>)`'s base implementation rents from `ArrayPool`, reads, copies and returns. `FileStream`/`MemoryStream` override it so the common path is unaffected, but there was no upside to the change either.

---

### 4. `ByteArrayExtensions.IsEntirelyNull` — public behaviour change

```csharp
public static bool IsEntirelyNull(byte[] buffer)
    => buffer.AsSpan().IndexOfAnyExcept((byte)0) == -1;
```

`((byte[])null).AsSpan()` is an empty span, so this now returns `true` for `null` where it previously threw `NullReferenceException`. This is a public method, so it's the one non-internal behaviour change in the PR. Probably an improvement, but worth making deliberate with `ArgumentNullException.ThrowIfNull` rather than leaving it implicit.

---

### 5. Nits

- `Mp3Frame.IsValidHeader` gained `if (headerBytes.Length != 4) return false;` with a commented-out `throw` above it (line 187, and a second at line 197). Both callers pass exactly 4 bytes, so the guard is unreachable — `Debug.Assert` or deletion would be cleaner than commented-out code.
- The whitespace re-alignment of the MP3 bitrate tables and one row of `MuLawDecompressTable` is unrelated to the change and makes the diff harder to read. Reverting those would help.
- The `XingHeader.sr_table` deletion is correct (nothing references it) but is unrelated dead-code removal; a separate commit would keep this PR focused.

Seb-stian:
@markheath Thank you for the comment.

1. Will replace with `ReadExactly`. I forgot it could return less bytes *even before* EOF.
2. Will experiment with the suggestions.
3. Missed that one.
4. **Q:** I assume we want the old behavior back?
5. I will ignore these, unless you want them addressed.

PR #1427, opened 2026-09-02
Fix winmm and WaveFormat interop under Native AOT (#1425)

### LINKED ISSUE #1425: WaveIn recording fails with 'The wave header was not prepared' in AOT build
Project to reproduce the bug: [NAudioBug2.zip](https://github.com/user-attachments/files/31654737/NAudioBug2.zip)

```
dotnet publish -c Release
```
then launch `bin\Release\net10.0-windows7.0\win-x64\native\Project.exe`
```
Unhandled exception. NAudio.MmException: WaveHeaderUnprepared calling waveInAddBuffer: The wave header was not prepared.  Use the Prepare function to prepare the header, and then try again.
   at NAudio.Wave.WaveInBuffer.Reuse() + 0x67
   at NAudio.Wave.WaveIn.DoRecording() + 0x44
   at NAudio.Wave.WaveIn.RecordThread() + 0x19
--- End of stack trace from previous location ---
   at Project.Program.<>c.<TestNAudio>b__1_1(Object, StoppedEventArgs) + 0x23
   at NAudio.Wave.WaveIn.RecordThread() + 0x31
   at System.Threading.Thread.StartThread(IntPtr) + 0x11f
   at System.Threading.Thread.ThreadEntryPoint(IntPtr) + 0x1a
```

---

Native interop itself is not an issue for AOT. Check the `TestCSWin32` function which invokes native methods via CsWin32. Recording goes well, output file is playable.

I guess there may be some subtle issue in current WaveInterop declarations. Or something getting GC'd too early maybe.

markheath:
Fixes #1425.

## Summary

`WaveIn` recording fails immediately in an AOT-published app with `WaveHeaderUnprepared calling waveInAddBuffer`. Investigating turned up **two** independent bugs, both from passing a `[StructLayout]` **class** to native code by value — something CoreCLR handles by pinning the managed object in place, and NativeAOT does not reproduce.

**1. `WAVEHDR` (the reported bug).** CoreCLR pins a blittable class argument and passes the address of the managed object, so winmm wrote `WHDR_PREPARED` / `WHDR_DONE` / `dwBytesRecorded` straight into it. NativeAOT copies the argument into a per-call stack temporary, In-only, so the driver's writes were discarded — `InQueue` and `Done` were permanently false and the first `waveInAddBuffer` returned `WAVERR_UNPREPARED`. A `WAVEHDR` also has to outlive the call that submits it, since the driver keeps the address queued. `WaveHeader` is now a struct in a stable `AllocHGlobal` block and the six header p/invokes take an `IntPtr` — the pattern `MidiIn` already used for `MIDIHDR`.

**2. `WAVEFORMATEX` (found while fixing the first).** NativeAOT drops the base-class fields when marshalling a class hierarchy, so `Marshal.SizeOf(waveFormatExtensible)` returned 22 instead of 40 and `MarshalToPtr` wrote the extensible fields at offset 0, over the sample rate. A round trip came back as `sampleRate=65536, bits=128`. This lives in `NAudio.Core`, so it affected **WASAPI, Media Foundation, DMO and ACM**, not just winmm. `MarshalToPtr`/`MarshalFromPtr` now build and parse the WAVEFORMATEX blob explicitly through the existing subclass-aware `Serialize`/`BinaryReader` pair, and every remaining site that sized or wrote a `WaveFormat` with `Marshal.SizeOf`/`StructureToPtr` was converted.

Also in here:

- `Mp3WaveFormat` gained the `Serialize` override it never had. It advertised `cbSize = 12` but wrote none of its 12 MPEGLAYER3WAVEFORMAT bytes — a malformed `fmt ` chunk on **every** runtime, the one genuinely non-AOT bug fixed here.
- `NAudio.WinMM` is marked `IsAotCompatible` and passes the trim/AOT analyzer clean.
- `WmaWaveFormat` and `OggWaveFormat` deleted — internal, zero references, and `NAudio.Core` exposes internals to no assembly.

### Breaking change

`AudioClient.IsFormatSupported` and `WasapiPlayer.IsFormatSupported` now return the closest-match format as **`out WaveFormat`** rather than `out WaveFormatExtensible`.

WASAPI documents `ppClosestMatch` as *"a **WAVEFORMATEX** or **WAVEFORMATEXTENSIBLE** structure"*, so the old signature could only represent half of what the API returns. It papered over that by reading 40 bytes from what may be an 18-byte `CoTaskMem` block: the base fields came out valid and the extensible fields were whatever followed the allocation. Decoding it honestly meant a plain format had nowhere to go, so widening the parameter is the fix rather than returning `null` and losing the suggestion.

Narrow in practice: per the API contract the closest match is **always NULL in exclusive mode**, and `WasapiOut` — the only consumer in the library — only asks for it there, using it in a null-guarded error message. Nothing in NAudio ever received a non-null value from this parameter. Callers wanting the extensible fields pattern-match.

`MarshalToPtr`'s allocation for `WaveFormatExtraData` also shrinks from a fixed 118 bytes to `18 + cbSize` — consistent with the advertised `cbSize`, so no conforming consumer can overread.

## Release notes

## Testing

**Automated.** Full solution builds clean. 1532 cross-platform tests pass, including 10 new ones in `WaveFormatMarshalTests` covering blob layout, round trips, the "blob is never shorter than its own cbSize" invariant across every shipped subclass, and the under-reported-`cbSize` path. The AOT smoke test gained a winmm phase whose blob assertions need no audio hardware, so they are a real regression guard anywhere; `TreatWarningsAsErrors` + `TrimmerSingleWarn=false` mean zero trim/AOT warnings from `NAudio.WinMM` is a meaningful signal.

**Empirical AOT verification.** Both bugs were reproduced and then confirmed fixed under genuine `PublishAot`, comparing JIT against AOT:

- A winmm-shaped native stub showed `waveInAddBuffer` returning `34` (`WAVERR_UNPREPARED`) with the driver's writes discarded before the fix, and identical JIT/AOT behaviour after.
- The format round trip across PCM / IEEE float / Extensible / ADPCM / GSM610 / MP3 / TrueSpeech is byte-identical between JIT and AOT after the fix, and was corrupt under AOT before.

**Reviewed.** A separate adversarial review pass proved the new blob is byte-for-byte identical to the old `Marshal.StructureToPtr` output on CoreCLR for all shipped subclasses — the point being that this is not a behaviour change on the JIT. It found five sites the first pass missed (DMO ×2, Media Foundation, ACM ×2), a regression where a subclass declaring `extraSize` without a `Serialize` override emitted a block shorter than its own `cbSize`, and a post-`Dispose` race that the native block escalated from a benign stale read into a use-after-free. All fixed in `7b484c0`.

**Not yet done — needs a Windows box.** No `PublishAot` link for `win-x64` was possible here, so ILC on Windows is unproven, and none of the runtime paths have been exercised against real audio hardware. `samples/NAudioConsoleTest/BatchPlans/winmm-waveformat-regression.json` is a ready batch plan covering the WAVEHDR path, ACM, WASAPI shared/exclusive, DMO and Media Foundation:

```
NAudioConsoleTest run-batch BatchPlans/winmm-waveformat-regression.json --out=results
```

Worth pointing it at a 24-bit or multichannel WAV — those are what travel as `WAVE_FORMAT_EXTENSIBLE`, and 16-bit 44.1kHz stereo would pass even with the format bug present.

---

Branch name is the auto-generated one rather than something descriptive as CLAUDE.md asks; happy to move it to `fix/1425-winmm-aot-marshalling` if you'd prefer.

https://claude.ai/code/session_01Wc2YvXR3bnMNBgCas7GsiN


markheath:
testing on windows complete

PR #1441, opened 2026-09-10
[NAudio 3] Fix XML documentation in most NAudio audio primitives

mdcdi1315:
## Summary

This PR aims to fix the poorly provided documentation on the `WaveStream`, `IWaveProvider` and `ISampleProvider` classes.

The `WaveProviders.md` document has been also updated to match the new Span API contract these classes do require from now on, and it's text is also updated to better describe what each class does.

As a side effect, the `IWavePlayer` XML docs have been also updated, and the extension interfaces (`IWavePosition` and `IWaveLatency`) have gained their own source file.

Feel free to push back on the provided XML documentation if you feel that there is missing something, or if something was written too verbose or anything else.

While modifying the XML docs in the `WaveStream` class, there was also something that captured my attention: the `CurrentTime` setter does not bound-check the Position value it passes to the Position setter. Typically, all the Position setters do bound-check so this is not a bug or a concerning issue but it is rather done to have it right.

By the way, I am also working on the macOS wrappers documentation alongside this one, I will open the PR when that is ready.

## Release notes

## Testing

No testing required, most of the changes are documentation,
and rather does and a reliability fix in the WaveStream CurrentTime property setter.

### DIFF
--- a/Docs/WaveProviders.md
+++ b/Docs/WaveProviders.md
@@ -1,50 +1,79 @@
-# WaveStream, IWaveProvider and ISampleProvider
+﻿# WaveStream, IWaveProvider and ISampleProvider
 
-When you play audio with NAudio or construct a playback graph, you are typically working with either `IWaveProvider` or `ISampleProvider` interface implementations. This article explains the three main base interfaces and classes you will encounter in NAudio and when you might use them.
+When you play audio with NAudio or construct a playback graph, you are typically working with either `IWaveProvider` or `ISampleProvider` interface implementations. 
+This article explains the three main base interfaces and classes you will encounter in NAudio and when you might use them.
 
-## WaveStream
+## `WaveStream`
 
-`WaveStream` was the first base class in NAudio, and inherits from `System.IO.Stream`. It represents a stream of audio data, and its format can be determined by looking at the `WaveFormat` property.
+`WaveStream` was the first base class in NAudio, inherits from `System.IO.Stream`,
+and implements `IWaveProvider` in newer versions.
+It represents a stream of audio data, and its format can be determined by querying the `WaveFormat` property.
+Typically, it is also seekable (repositionable), which it means that it is possible to get any sample of the stream at will.
 
-It supports reporting `Length` and `Position` and these are both measured in terms of bytes, not samples. `WaveStreams` can be repositioned (assuming the underlying implementation supports that), although care must often be taken to reposition to a multiple of the `BlockAlign` of the `WaveFormat`. For example if the wave stream produces 16 bit samples, you should always reposition to an even numbered byte position.
+Seeking (reposition) is implemented through the `System.IO.Stream` `Length` and `Position` properties. 
+These properties are both measured in terms of bytes, not samples.
+When a `WaveStream` is seekable (repositionable), care must often be taken to seek (reposition) to a multiple of the `BlockAlign` of the `WaveFormat`.
+If, for example, the stream produces 16 bit samples, you should always seek (reposition) to an even numbered byte position.
 
-Audio data is from a stream using the `Read` method which has the signature:
+Audio data are retrieved from a stream using the `Read` method which has the signature:
 
 ```c#
-int Read(byte[] destBuffer, int offset, int numBytes)
+int Read(Span<byte> buffer);
 ```
 
-This method is inherited from `System.IO.Stream`, and works in the standard way. The `destBuffer` is the buffer into which audio should be written. The `offset` parameter specifies where in the buffer to write audio to (this parameter is almost always 0), and the `numBytes` parameter is how many bytes of audio should be read.
+This method is also inherited from `System.IO.Stream`, and works in the standard way. 
+The `buffer` parameter is the buffer into which audio is expected to be written. 
+It's `Length` property denotes how many bytes are expected to be written to it.
 
-The `Read` method returns the number for bytes that were read. This should never be more than `numBytes` and can only be less if the end of the audio stream is reached. NAudio playback devices will stop playing when `Read` returns 0.
+The return value of the `Read` method must provide the number of bytes that were read. 
+This should never be more than `buffer.Length` bytes and can only be less if the end of the audio stream is reached. 
+By convention, NAudio playback devices will stop playing when `Read` returns 0.
 
-`WaveStream` is the base class for NAudio file reader classes such as `WaveFileReader`, `Mp3FileReader`, `AiffFileReader` and `MediaFoundationReader`. It is a good choice of base class because these inherently support repositioning. `RawSourceWaveStream` is also a `WaveStream`, and delegates repositioning requests down to its source stream.
+`WaveStream` is the base class for NAudio file reader classes such as `WaveFileReader`, `Mp3FileReader`, `AiffFileReader` and `MediaFoundationReader`.
+It is a good choice of base class because these inherently support repositioning. 
+`RawSourceWaveStream` is also a `WaveStream`, and delegates repositioning requests down to its source stream.
 
-For a more detailed look at all the methods on `WaveStream`, see [this article](http://markheath.net/post/naudio-wavestream-in-depth)
+For a more detailed look at all the methods on `WaveStream`, see [this article](http://markheath.net/post/naudio-wavestream-in-depth).
 
-## IWaveProvider
+## `IWaveProvider`
 
-Implementing `WaveStream` can be quite a lot of work, and for non-repositionable streams can seem like overkill. Also, streams that simply read from a source and modify or analyse audio as it passes through don't really benefit from inheriting from `WaveStream`.
+The `WaveStream` class is not suitable for all possible audio cases, and implementing all the properties/methods it requires can seem like overkill.
+Additionally, not every audio primitive provides immutable samples (samples that are not changed during the lifetime of the stream).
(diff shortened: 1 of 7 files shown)


src/NAudio.Core/Wave/WaveOutputs/ISampleProvider.cs
```diff
@@ -1,21 +1,65 @@
...
 
 namespace NAudio.Wave;
 
 /// <summary>
-/// Like IWaveProvider, but makes it much simpler to put together a 32 bit floating
-/// point mixing engine
+/// Like the <see cref="IWaveProvider"/> interface, the <see cref="ISampleProvider"/> interface is 
+/// also a low-level primitive that is an audio source, but all the audio samples are provided
+/// as 32-bit floating-point. <br />
+/// This was primarily useful back in .NET Framework when the Span API still not existed and reinterpreting
```
markheath: I wouldn't lead with this historical information and make it sound like `ISampleProvider` is not very important. I view `ISampleProvider` as one of the most useful base interfaces for people implementing anything involving sample manipulation or metering etc. The recommendation in NAudio is to get to `ISampleProvider` whenever you want to work directly with samples yourself so you don't need to worry about bit depth.
I suggest removing most of this intro and focusing more on why `ISampleProvider` matters today

mdcdi1315: Okay, will get it done.

src/NAudio.Core/Wave/WaveStreams/WaveStream.cs
```diff
@@ -78,79 +116,80 @@ public override long Seek(long offset, SeekOrigin origin)
...
+        long length = Length;
         long newPosition = Position + WaveFormat.AverageBytesPerSecond * seconds;
         if (newPosition > Length)
-            Position = Length;
-        else if (newPosition < 0)
-            Position = 0;
+            newPosition = Length;
+        else if (newPosition < 0L)
+            newPosition = 0L;
         else
```
markheath: looks to me like this else needs to go, or we're not actually changing position when you skip outside the valid region

mdcdi1315: Yes, but TimeSpan can inherently accept negative values, as far as I remember.
So, passing a negative span of time when on position 0 should yield a negative value, which will throw exception on a correctly implemented `Position` setter. Except if I am mistaken, so please do correct me.

src/NAudio.Core/Wave/WaveStreams/WaveStream.cs
```diff
@@ -78,79 +116,80 @@ public override long Seek(long offset, SeekOrigin origin)
...
+    /// Moves forwards or backwards the specified number of seconds in the wave stream. <br />
+    /// If the computed new position exceeds the length of the wave stream, or is less than 0,
+    /// the value is clamped in the range [0..length], before setting it as a new position value.
     /// </summary>
     /// <param name="seconds">Number of seconds to move, can be negative</param>
+    /// <exception cref="IOException">An I/O error occurred.</exception>
+    /// <exception cref="ObjectDisposedException">The wave stream is disposed.</exception>
     public void Skip(int seconds)
     {
+        long length = Length;
```
markheath, line 145: this isn't actually used

mdcdi1315: Okay... let me see.

mdcdi1315: Yes, this was not used. 
The intention was to avoid calling `Length` two times. 
Have already fixed it.

Docs/WaveProviders.md
```diff
@@ -54,27 +83,41 @@ public interface ISampleProvider
...
-`ISampleProvider` is a great base interface to inherit from if you are implementing any kind of audio effects. In the `Read` method you typically read from your source `ISampleProvider`, then modify the floating point samples, before returning them. Here's the implementation of the `Read` method in `VolumeSampleProvider` showing how simple this can be:
+To summarize, the `ISampleProvider` is a great base interface to use when you
+are implementing any kind of an audio effect.
+
+In the `Read` method you typically read from your source `ISampleProvider`, then modify the floating point samples, before returning them.
+Here's the implementation of the `Read` method in `VolumeSampleProvider` showing how simple this can be:
 
 ```c#
-public int Read(float[] buffer, int offset, int sampleCount)
+public int Read(Span<byte> buffer)
```
markheath: should be Span<float>

Docs/WaveProviders.md
markheath: also needs updating here `int Read(Span<float> buffer)`

src/NAudio.Core/Wave/WaveStreams/WaveStream.cs
mdcdi1315: Oh, this was a slipped coding error.

Docs/WaveProviders.md
```diff
@@ -54,27 +83,41 @@ public interface ISampleProvider
 }
 ```
 
-The `WaveFormat` will always be 32 bit floating point, but the number of channels or sample rate may of course vary.
+The `WaveFormat` will always be 32 bit floating point,
+but the number of channels or sample rate may of course vary by implementation.
+
+The `Read` method's `buffer` parameter specifies the buffer to place the processed samples into.
+The buffer's `Length` property denotes how many bytes are expected to be written to it.
```
markheath: samples, not bytes

src/NAudio.Core/Wave/WaveOutputs/ISampleProvider.cs
```diff
@@ -1,21 +1,65 @@
...
+/// samples as their numeric types needed loads of code in the Read implementation. For example, applying
+/// just a gain (volume) modifier on PCM 16 bit rate audio would require the following code:
```
markheath: bit depth, not bit rate

Docs/WaveProviders.md
mdcdi1315: Yes, that's correct!

src/NAudio.Core/Wave/WaveOutputs/ISampleProvider.cs
mdcdi1315: That's correct as well....

Docs/WaveProviders.md
```diff
@@ -1,50 +1,79 @@
...
+that can be played or rendered to a file.
 
 `BufferedWaveProvider` is a good example of a `IWaveProvider` as it has no ability to reposition - it simply returns any buffered audio from its `Read` method.
 
-## ISampleProvider
+## `ISampleProvider`
 
-The strength of `IWaveProvider` is that it can be used to represent audio in any format. It can be used for 16,24 or 32 bit PCM audio, and even for compressed audio (MP3, G.711 etc). But if you are performing any kind of signal processing or analysis on the audio, it is very likely that you want the audio to be in 32 bit IEEE floating point format. And it can be a pain to try to read floating point values out of a `byte[]` in C#.
+While implementing `IWaveProvider` is a useful approach to implement an audio effect, it has the caveat that
+decoding the samples depends of the bit rate (number of bits per sample) declared on the `WaveFormat`.
```
markheath: also two places here - bit depth, not bit rate

Docs/WaveProviders.md
mdcdi1315: will fix it...

Docs/WaveProviders.md
mdcdi1315: Yep...

src/NAudio.Core/Wave/WaveStreams/WaveStream.cs
```diff
@@ -78,79 +116,80 @@ public override long Seek(long offset, SeekOrigin origin)
...
     public virtual TimeSpan CurrentTime
     {
-        get
-        {
-            return TimeSpan.FromSeconds((double)Position / WaveFormat.AverageBytesPerSecond);
-        }
+        get => TimeSpan.FromSeconds((double)Position / WaveFormat.AverageBytesPerSecond);
         set
         {
+            long length = Length;
```
markheath: this is a small behavioural change (clamps), and also requires Length not to throw, so if we do this change we probably need a brief note in release notes mentioning the change.

mdcdi1315: Yes of course, but clamping is better than not clamping - because there can be cases that clients just add through the CurrentTime property 10 seconds repeatedly, without doing a bound check that they can move to the next 10 seconds, just like Skip does today: check before seeking. So either the stream is ending up outside the audio data stream (does not throw but does return non-audible data from there, if assuming a fully managed reader) or even worse throwing argument out of range because the position ended up outside the stream's bound.

But about the length, I will add a try handler to catch any exceptions from it, and will ensure to be used only when the length is actually defined.

markheath: yes, clamping is better, and  length not supported is an edge case so not super important (really should be using IWaveProvider for infinite streams rather than WaveStream but I guess could be possible in some streaming situation which would also likely not be seekable)

mdcdi1315:
Sorry for pushing that 1 hour later - I was experimenting to make the ISampleProvider XML doc to have sense.

markheath:
Cool, I'll try to re-review tomorrow

Issue #672, opened 2020-07-22
WasapiLoopbackCapture raises unhandled exception when i change default audio devices


varz29:
Hi,
When I use WasapiLoopbackCapture  i get an unhandled exception when i try to change default audio out device ,
How can I handle this exception ?

markheath:
Look at the `Exception` property on the `RecordingStopped` event

varz29:
Its not raising an event on recording stopped.
Application goes into break state and i get below message

Managed Debugging Assistant 'FatalExecutionEngineError' 
  Message=Managed Debugging Assistant 'FatalExecutionEngineError' : 'The runtime has encountered a fatal error. The address of the error was at 0x11484265, on thread 0x19ec. The error code is 0xc0000005. This error may be a bug in the CLR or in the unsafe or non-verifiable portions of user code. Common sources of this bug include user marshaling errors for COM-interop or PInvoke, which may corrupt the stack.'


Managed Debugging Assistant 'FatalExecutionEngineError' 
  Message=Managed Debugging Assistant 'FatalExecutionEngineError' : 'The runtime has encountered a fatal error. The address of the error was at 0x11484265, on thread 0x19ec. The error code is 0xc0000005. This error may be a bug in the CLR or in the unsafe or non-verifiable portions of user code. Common sources of this bug include user marshaling errors for COM-interop or PInvoke, which may corrupt the stack.'

varz29:
This happens when the recording is inprogress

markheath:
What happens if you run in release mode (i.e. not debugging from Visual Studio)?

varz29:
It is the same with the release build also

varunistar:
Hi @markheath  is there any way to handle this exception.
Thanks

markheath:
I've made some improvements in this area. Previously when the device was removed we'd get a second exception trying to Stop and Reset recording which meant you couldn't catch the `RecordingStopped` event. That's been fixed for `WasapiCapture` (and `WasapiLoopbackCapture`) as well as `WasapiRecorder` (which is the replacement for both in NAudio 3).

Also for NAudio 3, `WasapiRecorder` supports automatic stream routing, so in theory you can just follow the default device (although I haven't tried that yet for loopback recording).

PR #1443, opened 2026-09-10
Fix WasapiPlayer left stranded when the source's Read throws (#1442)

### LINKED ISSUE #1442: NAudio 3: `WasapiPlayer` can be left into an invalid state if the first ever `IWaveProvider.Read` call throws
While I was trying to adjust the gain of an audio file (because it was too low), I decided to play it back to verify that that was the gain I wanted.
So, I injected a `VolumeSampleProvider` through the Media Foundation reader, changed the `Volume` property to the desired value and instructed the player to do playback.

Because I was using C# interactive to do this, I forgot an important step: To bring a reference to the `System.Numerics.Tensors` package, that the `VolumeSampleProvider` depends on. So, the result was that the `VolumeSampleProvider` `Read` call threw that the Tensors assembly was not found, and left the player into an invalid state, which it's `PlaybackState` reported `Playing`. Even after managing to provide the reference through the dynamic resolver, (and doing `Stop`, then `Play` due to it's already invalid state) the player said that the buffer was too large. The `PlaybackStopped` event normally dispatched and indicated the exception, however.

This is probably one of the bugs I have found by accident!

The exact code that I used:

~~~C#
// Add #r references as required, if using C# interactive.

using NAudio.Wave;
using NAudio.Wave.SampleProviders;

var reader = new MediaFoundationReader("<Path_To_File>");
var volumectl = new VolumeSampleProvider(reader.ToSampleProvider());

var player = new WasapiPlayerBuilder().Build();
player.PlaybackStopped += (_, e) => System.Console.WriteLine("Stopped: {0}", e.Exception);
player.Init(volumectl);
player.Play();

// If we forget to add the package reference, we get System.IO.FileNotFoundException through 
// PlaybackStopped, but the PlaybackState still reports Playing, but we should have stopped!
// Even after recovery, WASAPI reports that the buffer is too large!
~~~

The error stack trace (the buffer too large error):

~~~
Stopped: NAudio.CoreAudioApi.CoreAudioException (0x88890006): The buffer is too large.
   at NAudio.CoreAudioApi.CoreAudioException.ThrowIfFailed(Int32 hresult) in E:\.projects\NAudio\src\NAudio.Wasapi\CoreAudioApi\CoreAudioException.cs:line 35
   at NAudio.CoreAudioApi.AudioRenderClient.GetBufferLease(Int32 numFramesRequested, Int32 bytesPerFrame) in E:\.projects\NAudio\src\NAudio.Wasapi\CoreAudioApi\AudioRenderClient.cs:line 49
   at NAudio.Wave.WasapiPlayer.FillBuffer(Int32 frameCount) in E:\.projects\NAudio\src\NAudio.Wasapi\WasapiPlayer.cs:line 823
   at NAudio.Wave.WasapiPlayer.PlayThread() in E:\.projects\NAudio\src\NAudio.Wasapi\WasapiPlayer.cs:line 759
~~~

markheath:
Fixes #1442.

## Summary

The reported symptom was two independent bugs, both triggered by an exception escaping `IWaveProvider.Read`. (The `FileNotFoundException` in the issue is not itself an NAudio bug — `System.Numerics.Tensors` is a proper `PackageReference` on `NAudio.Core`, so NuGet consumers get it transitively; C# interactive with manual `#r` just doesn't resolve transitive dependencies. It's simply a convenient way to make `Read` throw.)

**1. "The buffer is too large" — `RenderBufferLease.Dispose` committed the buffer instead of discarding it.**

`FillBuffer` holds a WASAPI buffer lease across the `Read` call. When `Read` threw, `using` disposed the lease, and dispose defaulted to `framesWritten ?? frameCount` — i.e. it told WASAPI the *entire* buffer had been written, with undefined contents. Padding became the full buffer and stayed that way, so the next `Play()` asked `GetBuffer` for `bufferFrameCount` frames with nothing free and failed with `AUDCLNT_E_BUFFER_TOO_LARGE (0x88890006)` — exactly the stack trace in the issue.

`Dispose` now releases zero frames, which discards the packet, and swallows any failure so a dispose running while an exception is in flight can't replace the caller's exception with a less useful one. `Release(n)` is unchanged, and `FillBuffer` always calls it explicitly on success (nulling `owner`), so only the failure path behaves differently. `CaptureBufferLease.Dispose` is deliberately left alone — releasing the full frame count is the correct capture contract.

This is a NAudio 3 regression from the zero-copy design: legacy `WasapiOut.FillBuffer` reads into an intermediate managed buffer *before* calling `GetBuffer`, so a throwing provider never leaves a WASAPI buffer checked out.

**2. `PlaybackState` stayed `Playing` after the play thread died.**

`WasapiPlayer.PlayThread` only set `PlaybackState.Stopped` on the orderly end-of-stream path, and only stopped/reset the audio client there too. Two paths bypassed it: an exception from `Read`, and the `return` taken when the source ends before the first buffer is filled. Either left the player reporting `Playing` with no thread behind it.

Teardown moves into the `finally` block, using a `SafeStopAndReset()` helper that mirrors the one `WasapiRecorder.CaptureThread` already uses (added for #672). State is set before `RaisePlaybackStopped` so handlers observe `Stopped`.

**3. Same state bug in legacy `WasapiOut`**, fixed the same way for consistency. It was never exposed to bug 1, but it had the identical stranded-state defect, including the zero-length-stream early return.

One deliberate behaviour change worth flagging: `audioClient.Stop()` on the orderly end-of-stream path is now best-effort rather than surfaced through `PlaybackStopped`. That's the same tradeoff `WasapiRecorder` already makes, and device loss mid-playback still surfaces — `CurrentPadding`/`FillBuffer` in the loop throws first.

## Release notes

## Testing

`NAudio.Wasapi` and `NAudio.Windows.Tests` both compile clean with 0 warnings, and the 1542 cross-platform `NAudio.Core.Tests` pass (1539 passed, 3 skipped).

New in `tests/NAudio.Windows.Tests/Wasapi/WasapiPlayerRecoveryTests.cs`, covering both play-thread exit paths against a real render endpoint:

- `SourceThrowingOnFirstRead_StopsAndCanPlayAgain` — asserts the player reports `Stopped` after the failure and that a second `Play()` succeeds. That second `Play()` is the `AUDCLNT_E_BUFFER_TOO_LARGE` regression.
- `SourceEndingBeforeFirstBuffer_ReportsStopped` — the early-return path, before the audio client is ever started.

Both are `[Category("IntegrationTest")]` and `[Platform("Win")]`, so the headless CI run filters them out, and they `Assert.Ignore` when there's no default render endpoint.

**Verified on hardware:** both integration tests pass on a Windows machine with a real render endpoint, and the tone is audible after the recovery `Play()` — the path that previously failed with `AUDCLNT_E_BUFFER_TOO_LARGE`. The diagnosis also matches the reporter's stack trace exactly: `WasapiPlayer.cs:823` reached from `PlayThread:759` is the initial `FillBuffer` → `GetBufferLease`.

https://claude.ai/code/session_012N6xNEbsxFXMrjTERGsJg9

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -13,6 +13,7 @@ the NuGet `PackageReleaseNotes` field, which has a hard 35,000-character
 limit and fails the release build if exceeded.
 -->
 
+ * Fixed `WasapiPlayer` and `WasapiOut` being stranded in the `Playing` state, unable to play again, when the source's `Read` threw or the source ended before the first buffer was filled (#1442)
  * **macOS support (preview)** - A new wrappers library is added wrapping macOS native API's for playback/recording, reading/writing files and resampling audio. The `NAudio.MacOS` package ships pre-release only (`X.Y.Z-preview.N`) while its API settles, so it stays pre-release alongside stable releases of the other packages. Special thanks to @mdcdi1315 (#1325) for the work and the tests. For more information, see the [design](Docs/Architecture/MacOSWrappersDesign.md) doc that describes the decisions that helped shape the wrappers. (#1398)
 
 ### 3.1.0 (7 Sep 2026)
--- a/src/NAudio.Wasapi/CoreAudioApi/AudioRenderClient.cs
+++ b/src/NAudio.Wasapi/CoreAudioApi/AudioRenderClient.cs
@@ -114,10 +114,27 @@ public ref struct RenderBufferLease
     }
 
     /// <summary>
-    /// Releases the buffer with the full requested frame count.
+    /// Releases the buffer, discarding its contents if <see cref="Release(int?, AudioClientBufferFlags)"/>
+    /// was never called.
     /// </summary>
+    /// <remarks>
+    /// An unreleased lease reaching here means the caller never wrote any audio - typically because an
+    /// exception escaped the <c>using</c> block. Releasing the full frame count in that case would hand
+    /// WASAPI a buffer of undefined data and leave the stream padded full, so the next
+    /// <see cref="AudioRenderClient.GetBuffer"/> fails with AUDCLNT_E_BUFFER_TOO_LARGE (issue #1442);
+    /// releasing zero frames discards the packet instead. Never throws: a dispose running while an
+    /// exception is in flight must not replace it with a less useful one.
+    /// </remarks>
     public void Dispose()
     {
-        Release();
+        try
+        {
+            owner?.ReleaseBuffer(0, AudioClientBufferFlags.None);
+        }
+        catch
+        {
+            // The buffer is being abandoned; failing to hand it back is not actionable here.
+        }
+        owner = null;
     }
 }
--- a/src/NAudio.Wasapi/WasapiOut.cs
+++ b/src/NAudio.Wasapi/WasapiOut.cs
@@ -168,10 +168,6 @@ public class WasapiOut : IWavePlayer, IWavePosition, IWaveLatency
                 // behavior of waiting briefly for the last buffer to play
                 Thread.Sleep(isUsingEventSync ? latencyMilliseconds : latencyMilliseconds / 2);
             }
-            audioClient.Stop();
-            // set if we got here by reaching the end
-            playbackState = PlaybackState.Stopped;
-            audioClient.Reset();
         }
         catch (Exception e)
         {
@@ -179,8 +175,24 @@ public class WasapiOut : IWavePlayer, IWavePosition, IWaveLatency
         }
         finally
         {
+            // Always stop, reset and report Stopped, however the thread left the loop - reaching the
+            // end of the source, an exception from Read, or the zero-length-stream early return above.
+            // Otherwise PlaybackState stays Playing with no thread behind it (issue #1442).
+            SafeStopAndReset();
+            playbackState = PlaybackState.Stopped;
             RaisePlaybackStopped(exception);
         }
+    }
+
+    /// <summary>
+    /// Best-effort stop and reset of the audio client during teardown. A device that has been
+    /// removed mid-playback fails these calls (AUDCLNT_E_DEVICE_INVALIDATED); the failure is not
+    /// actionable here, and must not mask the real exception or escape and kill the play thread.
+    /// </summary>
+    private void SafeStopAndReset()
+    {
+        try { audioClient?.Stop(); } catch { /* device already gone */ }
+        try { audioClient?.Reset(); } catch { /* device already gone */ }
     }
 
     private void RaisePlaybackStopped(Exception e)
--- a/src/NAudio.Wasapi/WasapiPlayer.cs
+++ b/src/NAudio.Wasapi/WasapiPlayer.cs
@@ -797,10 +797,6 @@ public class WasapiPlayer : IWavePlayer, IWavePosition, IWaveLatency, IAsyncDisp
                     }
                 }
             }
-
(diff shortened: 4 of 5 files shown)


markheath:
CI note: the `build` check is red on one test, and I don't believe it belongs to this PR.

**Failing:** `EffectAllocationTests.SteadyStateProcessDoesNotAllocate(NAudio.Effects.CompressorEffect)` — "CompressorEffect allocated 1424 bytes across 512 Process calls" (`tests/NAudio.Core.Tests/Effects/EffectAllocationTests.cs:50`). `NAudio.Windows.Tests` passed, including the new WASAPI integration tests being correctly filtered out of the headless run.

**Why it isn't this PR's:** every production change here is in `NAudio.Wasapi`, and `NAudio.Core.Tests` references only `NAudio.Core` and `NAudio.Midi` — it cannot load the changed assembly. `CompressorEffect.ProcessBlock` is also allocation-free by inspection: pure float math over a `Span<float>`, no boxing, closures or lazy buffers on the steady-state path.

**Likely cause:** the test warms up with 64 `Process` calls before measuring, which sits right on .NET's tier-1 promotion threshold (~30 calls plus a background delay). On a slow or loaded runner the re-JIT/OSR transition can land inside the measured window, and `GC.GetAllocatedBytesForCurrentThread()` attributes that to the calling thread. 1424 bytes as a one-off, rather than a per-call cost, fits that. It would also explain why `CompressorEffect` is the one that trips — it has the heaviest inner loop of the six representatives.

**No fix exists to port,** and `main` is green across its last 10 runs, so this looks like a latent flake rather than a regression. I don't have permission to re-run jobs on this repo, so I can't confirm it that way. I've deliberately not touched the test — it's unrelated to this change, and I'd rather not widen the PR or weaken an assertion that is pinning a real guarantee.

If it recurs, the targeted fix is to make the warm-up JIT-stable rather than relax the assertion — raise the warm-up count well past the tier-1 threshold, and/or call `RuntimeHelpers.PrepareMethod` on the `Process` path before measuring. Happy to do that here or as a separate PR, whichever you prefer.

---
_Generated by [Claude Code](https://claude.ai/code)_

markheath:
Correcting my note above on two points, having tried to reproduce it.

**The tiered-JIT explanation is unconfirmed.** It did not reproduce in ~5,500 samples: a standalone probe replicating the test (one process per cold-JIT opportunity) gave zero dirty first windows across 400 idle runs, 200 runs at 3× CPU oversubscription, and a sweep of `DOTNET_TC_CallCountingDelayMs` from 0 to 400 ms; looping the real test 40 times under load added 240 more clean samples. So on Linux/Release the rate is below roughly 1 in 5,000 case-executions. I stated that mechanism more confidently than the evidence supports — the "one-off rather than per-call" reading still holds (1424 bytes over 512 calls is ~2.8 bytes/call, and `ProcessBlock` has no allocating path), but the specific cause does not.

**`RuntimeHelpers.PrepareMethod` was a bad suggestion** — it guarantees the method is compiled, but under tiered compilation it can still be tier-0 and re-JIT later, which is precisely the event it was meant to exclude. Disregard that half.

The fix is now naudio/NAudio#1444, deliberately mechanism-agnostic: measure up to three consecutive windows and stop at the first clean one, so a per-call allocation still fails every window while a one-off cannot fail the run. Verified by injecting both shapes of failure into the real test method. It leaves this PR untouched.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1398, opened 2026-07-29
macOS wrappers PR for NAudio 3

mdcdi1315:
## Summary

This is the PR bringing macOS wrappers to NAudio 3.

See issue #1325 for the original issue, research and intent.

There is still work to do until this gets merged in, I just have uploaded it as PR here so that reviews can be done.

## Release notes

- [x] I've added a one-line bullet to the `### Unreleased` section of [RELEASE_NOTES.md](../RELEASE_NOTES.md)
- [ ] This PR doesn't need a release-notes entry (internal refactor, test-only, or docs fix)
- [ ] Leave the release-notes entry for the maintainer to write

## Testing

There is a tests project that performs various tests about the reliability of the wrappers - see `NAudio.MacOS.Tests` project.

Tests performed against the commit [4934df7](https://github.com/naudio/NAudio/pull/1398/commits/4934df726c4ebe62d664e47ebe3d7195e8aead91):

~~~
total: 179
failed: 0
succeeded: 179
skipped: 0
~~~

PR #1447, opened 2026-09-11
macOS wrappers documentation

mdcdi1315:
## Summary

Follow-up PR for the PR #1398.

This PR covers the documentation part of the original macOS wrappers PR, which that PR did not shipped.

It provides a full set of documentation for all the new API's provided by the macOS wrappers.
The TOC has also been updated.

Some API's were also renamed during this process so that the intent is better understood.

Additionally, the `Resampling.md` document has been updated to add the macOS resampler into the list of available resamplers to use.

## Release notes

## Testing

No testing required, this just updates the documentation.

### DIFF
--- a/Docs/Resampling.md
+++ b/Docs/Resampling.md
@@ -1,33 +1,74 @@
-# Resampling Audio
+﻿# Resampling Audio
 
 Every now and then you'll find you need to resample audio with NAudio. For example, to mix files together of different sample rates, you need to get them all to a common sample rate first. Or if you're playing audio through an API like ASIO, audio must be resampled to match the output device's current sample rate before being sent to the device.
 
 There are also some gotchas you need to be aware of when resampling. In particular there is the danger of "aliasing". The main takeaway is that if you lower the sample rate, you really ought to use a low pass filter first, to get rid of high frequencies that cannot be correctly represented at the lower sample rate.
 
-## Option 1: MediaFoundationResampler
+## Option 1: MediaFoundationResampler (Windows)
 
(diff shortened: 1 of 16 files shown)


mdcdi1315:
Remaining items to update: The lndex page in API reference, the repo's README, and the assembly layout plan, but these are rather tasks that I do not know what you want to write to them, so I am leaving that for you.

Docs/Resampling.md
```diff
@@ -1,33 +1,74 @@
...
 
-## Option 1: MediaFoundationResampler
+## Option 1: MediaFoundationResampler (Windows)
 
 Probably the most powerful resampler available with NAudio is the `MediaFoundationResampler`. This uses the Windows Media Foundation resampler MFT, which is available on all supported versions of Windows (Windows 10+). It has a customisable quality level (60 is the highest quality, down to 1 which is linear interpolation). It's fast enough to run on top quality. It's also quite flexible and is often able to change to a different channel count or bit depth at the same time.
 
 Here's a code sample that resamples an MP3 file (usually 44.1kHz) down to 16kHz. The `MediaFoundationResampler` takes an `IWaveProvider` as input, and a desired output `WaveFormat`:
 
-```c#
+~~~c#
```
markheath: any reason for switching from backticks (which we use everywhere in docs) to tildes?

mdcdi1315: To distinguish it from the single-line code (that is using backticks). If you want to, I can switch it back to the original backticks.

markheath: Yes, not a big deal, but the standard for all the docs is backticks for code blocks, so I'd like to keep it that way. Apart from that this is good to go in

mdcdi1315: OK, will fix that with a commit later on today.

mdcdi1315:
Updated the docs now. 
Please ensure first that this is what you want to, before merging this.
I have also updated a bit the text in the recording permissions section (in the recording document).
If there is anything else that you feel that is wrong, feel free to comment.

PR #1455, opened 2026-09-19
Refactor AIFF

Seb-stian:
## Summary

- Add `Span` overloads for `AiffFileWriter.Write` and `AiffFileReader.Read`
- Replace hand-rolled frexp/ldexp in `IEEE` with `Math.ILogB/Math.ScaleB`
- Replace manual big-endian byte packing/unpacking with `BinaryPrimitives`
- Avoid per-chunk string allocations in `AiffFileReader` by reading chunk names into an inline-array `ChunkName` struct
- Reduce `lock` scope in `AiffFileReader.Read` to not include the thread-safe portion
- Use stackalloc/ArrayPool<byte> in `AiffFileWriter.Write` to avoid a full-buffer heap allocation on every write
- Generic refactor for simpler code

Got rid of many unnecessary `byte[]` and `string` allocations! 🎉

## Release notes

- [X] Leave the release-notes entry for the maintainer to write

Possibly document the added public methods.

## Testing

All tests passing. Claude Sonnet 5 review.


src/NAudio.Core/Utils/IEEE.cs
```diff
@@ -12,23 +12,28 @@ public static class IEEE
...
     {
-        return ((long)(u - 2147483647L - 1) + 2147483648.0);
+        return u;
     }
 
-    private static double ldexp(double x, int exp)
+    private static double Ldexp(double x, int exp)
     {
-        return x * Math.Pow(2, exp);
+        return Math.ScaleB(x, exp);
```
mdcdi1315: Suggestion: Maybe we should add an AggressiveInlining hint for these two methods. (`UnsignedToFloat` and `Ldexp`) These two just execute intrinsic methods or do low-level operations. Would help the JIT.

The same goes for the `FloatToUnsigned` method.

mdcdi1315:
A question: Should we use `BinaryWriter` and `BinaryReader` anymore? It is little-endian only (as far as I remember, correct me if I am wrong) and `BinaryPrimitives` can fully replace the reader/writer. Sure I want to provide to the library at some time a static class with streaming operations that read and write primitives and string reading/writing, but maybe for now we should leave from the `BinaryReader`/`BinaryWriter` API and explicitly handle the endianness for each number written to a stream.

By the time of writing this, I have also checked the WAV reader/writer sources, seems like they need code updates too.

src/NAudio.Core/Utils/IEEE.cs
Seb-stian: Probably won't make a difference, but why not.

Seb-stian:
> It is little-endian only

Correct.

> BinaryPrimitives can fully replace the reader/writer

Can something replace `Read(Span<char>)`? That's the one I had to keep `BinaryReader` for.

> the WAV reader/writer sources, seems like they need code updates too

I can take a stab at it.

mdcdi1315:
> Read(Span&lt;char&gt;)

Since the BinaryReader uses UTF-8 by default (which is that it is used on the AIFF reader because the single-arg ctor is called), it could be replaced with UTF-8 strings. The chunk names in most formats, including AIFF, are written with ASCII so it would be compatible.

Seb-stian:
> with UTF-8 strings

With the hard assumption that 1 char = 1 byte? Because otherwise we'd have to handle larger code units.

I *think* that the assumption is correct, but I'm asking just in case.

mdcdi1315:
> I think that the assumption is correct, but I'm asking just in case.

Yes it is correct since chunk names are four byte codes, so they are ASCII anyway. But allow me to check a bit on AIFF and will comment back.

mdcdi1315:
Found some old specifications by Apple for this, they say that this is the format of a chunk ID: 

```
ID: 32 bits, the concatenation of four printable ASCII character in the range '
' (SP, 0x20) through '~' (0x7E). Spaces (0x20) cannot precede printing
characters; trailing spaces are allowed. Control characters are forbidden.
```

So it is correct to assume UTF-8 strings.

Seb-stian:
Changed `ChunkName` to bytes and removed `BinaryReader`.

src/NAudio.Core/Utils/IEEE.cs
```diff
@@ -10,25 +11,33 @@ namespace NAudio.Utils;
 public static class IEEE
 {
     #region Helper Methods
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
     private static double UnsignedToFloat(ulong u)
     {
-        return ((long)(u - 2147483647L - 1) + 2147483648.0);
+        return u;
```
markheath: what is the point of this method existing now if it just collapses into a cast?

src/NAudio.Core/Utils/IEEE.cs
```diff
@@ -10,25 +11,33 @@ namespace NAudio.Utils;
...
+        }
+        exp = Math.ILogB(x) + 1;
+        return Math.ScaleB(x, -exp);
     }
 
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
     private static ulong FloatToUnsigned(double f)
     {
-        return ((ulong)(((long)(f - 2147483648.0)) + 2147483647L) + 1);
+        return (ulong)f;
```
markheath: Same here - do we even need this method to exist anymore?

markheath:
Were you able to play an AIFF file successfully and confirm the audio sounds OK after this change?
The code changes seem OK to me, but the IEEE modifications would best be proved by converting a WAV to AIFF with AiffFileWriter and then playing it with AiffFileReader.

Seb-stian:
> Were you able to play an AIFF file successfully and confirm the audio sounds OK after this change?

Tested on https://archive.org/details/hanna-barbera-cartoon-sound-fx-fixed-album/01+Hanna-Barberas+Top+Ten+Famous+Sound+Effects.wav

```cs
using var wavReader = new WaveFileReader(@"path\01 Hanna-Barberas Top Ten Famous Sound Effects.wav");
AiffFileWriter.CreateAiffFile(@"path\01 Hanna-Barberas Top Ten Famous Sound Effects.aiff", wavReader);

using var aiffReader = new AiffFileReader(@"path\01 Hanna-Barberas Top Ten Famous Sound Effects.aiff");
using var output = new WasapiPlayerBuilder().Build();
output.Init(aiffReader);
output.Play();
while (output.PlaybackState == PlaybackState.Playing)
{
    Thread.Sleep(1000);
}
```

Also tried playing with VLC. Sounds good to me.

Issue #1432, opened 2026-09-04
[bug]: WaveFormatExtensible is silently corrupted under NativeAOT


diamondStar35:
## Summary

Under NativeAOT, `Marshal.SizeOf<WaveFormatExtensible>()` returns **22** instead of **40**, because
NativeAOT counts only the fields the derived class declares and ignores the 18 bytes inherited from
`WaveFormat`. Every `Marshal.PtrToStructure<WaveFormatExtensible>` therefore reads the extension fields
from offset 0, and the base fields are never populated at all — they keep their constructor defaults.

The practical consequence is that `IAudioClient::GetMixFormat` — reached through
`MMDevice.CreateAudioClient().MixFormat`, and used internally by `WasapiRecorderBuilder` when no explicit
format is requested — returns a `WaveFormat` describing a device that does not exist. Passing it back to
`IAudioClient::Initialize`, which is what NAudio does by default, then fails:

```
NAudio.CoreAudioApi.AudioFormatNotSupportedException (0x88890008):
The audio format is not supported by the audio endpoint device.
```

**A caller cannot avoid this from outside NAudio**, because the corruption happens before the format
object is handed back. `WasapiRecorderBuilder().WithDevice(device).Build()` — the documented minimal
capture setup — cannot work in an AOT build on any device whose mix format is `WAVE_FORMAT_EXTENSIBLE`,
which in practice is every modern endpoint.

## Environment

| | |
|---|---|
| OS | Windows 11 Pro, `DisplayVersion` 25H2, build **26200.9278**, x64 |
| `Environment.OSVersion` | `Microsoft Windows NT 10.0.26200.0` |
| .NET SDK | 10.0.111 |
| Package | `NAudio.Wasapi` 3.0.1 |

## Reproduction

The first half needs no audio hardware: it marshals a byte-for-byte correct `WAVEFORMATEXTENSIBLE` built
by hand. Run the program twice — once published with `PublishAot=true`, once under the runtime — and
compare the two outputs.

`repro.csproj`

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net10.0-windows10.0.19041.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
    <PublishAot>true</PublishAot>
    <RuntimeIdentifier>win-x64</RuntimeIdentifier>
  </PropertyGroup>
[...]

PR #1456, opened 2026-09-20
DirectSoundOut: marshal the secondary buffer's WAVEFORMATEX via WaveFormat.MarshalToPtr (regression since 3.1.0)

wolfdate25:
### Summary

`DirectSoundOut` fails every playback on NAudio 3.1.0 with `DirectSoundException 0x80004001` (`E_NOTIMPL`), thrown from `InitializeDirectSound` on the playback thread. `Init`/`Play` succeed, then `PlaybackStopped` fires almost immediately. Every source format is affected — a plain `new WaveFormat(44100, 16, 2)` included — on any device, including the default (`DSDEVID_DefaultPlayback`). The same code works on 2.3.0 on the same machine.

### Root cause

`DirectSoundOut.InitializeDirectSound` was the **last remaining site still passing a GCHandle-pinned `WaveFormat` instance to native code** as the `WAVEFORMATEX` for `CreateSoundBuffer`. The 3.1.0 Native-AOT work (#1432) removed `[StructLayout]` from `WaveFormat`, leaving the class with `LayoutKind.Auto` — so the pinned instance's memory is field-reordered and no longer a valid `WAVEFORMATEX`:

```
new WaveFormat(44100, 16, 2), pinned, first 18 bytes:

3.1.0:  44-AC-00-00 10-B1-02-00 01-00 02-00 04-00 10-00 00-00   → tag=44100, ch=0, rate=176400
2.3.0:  01-00 02-00 44-AC-00-00 10-B1-02-00 04-00 10-00 00-00   → tag=1, ch=2, rate=44100
expected (WAVEFORMATEX PCM16): 01-00 02-00 44-AC-00-00 ... 04-00 10-00 ...
```

`CreateSoundBuffer` reads that garbage format and returns `0x80004001`.

There is a second failure mode in the same line: a `WaveFormatExtraData` (which is what `WaveFileReader` returns for a WAV in 3.x) holds a `byte[] extraData` reference field, so the old `GCHandle.Alloc(waveFormat, GCHandleType.Pinned)` throws `ArgumentException: Object contains references` before the buffer is even created. This half also reproduces on 2.3.0 — playing a WAV file through `DirectSoundOut` has been broken by the same pin for longer.

Verified that neither the native API nor the new `[GeneratedComInterface]` projection is at fault: a raw P/Invoke replication of the full `InitializeDirectSound` sequence (DirectSoundCreate → SetCooperativeLevel → CreateSoundBuffer primary → Play → CreateSoundBuffer secondary → GetCaps → QI IDirectSoundNotify → SetNotificationPositions → Play) returns `S_OK` on every step, as does a verbatim replica of the interface definitions driven through `StrategyBasedComWrappers`.

### Fix

Replace the pinned `GCHandle` with `WaveFormat.MarshalToPtr` / `Marshal.FreeHGlobal` — the exact pattern every other converted site uses (`AudioClient.Initialize`, `WaveOutWindow`, `AcmDriver`, `MediaFoundationHelpers`). Fixes both failure modes.

### Verification (real hardware, Windows, .NET 10)

Same three-source matrix against the released 3.1.0 packages and against this branch:

| Source | NAudio 3.1.0 | This branch |
|---|---|---|
| `RawSourceWaveStream` + `new WaveFormat(44100,16,2)` | dies, `DirectSoundException 0x80004001` | alive, `GetPosition` advances |
| `RawSourceWaveStream` + `CreateIeeeFloatWaveFormat(44100,2)` | dies, `DirectSoundException 0x80004001` | alive, `GetPosition` advances |
| `WaveFileReader` (→ `WaveFormatExtraData`) | dies, `ArgumentException: Object contains references` | alive, `GetPosition` advances |

`NAudio.Core.Tests` with `TestCategory!=IntegrationTest`: 1557 passed, 0 failed. The new DirectSound integration tests (below) pass on a machine with real audio hardware and are marked `[Category("IntegrationTest")]` so the headless CI run skips them.

### Tests added

`DirectSoundTests` gained two integration tests: playing silence in a plain 16-bit PCM format, and playing a WAV through `WaveFileReader` (asserting the format really is a `WaveFormatExtraData`), each asserting `GetPosition` advances and `PlaybackStopped` never fired with an exception. Both fail with `0x80004001` / `ArgumentException` on 3.1.0.

### Release notes

`RELEASE_NOTES.md` `### Unreleased` bullet added per CLAUDE.md.


markheath:
Thanks for reporting, and a shame this got missed on the prior fix - DirectSoundOut is not a part of NAudio I use very much. I'll do a test locally and then get this in .

